Issue #6, opened 2017-04-22
Plugin fails on missing kotlin-gradle-plugin dependency


jeremymailen:
with `build.gradle`:
```groovy
plugins {
    id 'org.jetbrains.kotlin.jvm' version '1.1.1'
    id 'org.jlleitschuh.gradle.ktlint' version '1.0.2'
}
```
`gradle tasks` fails with:
```
> Could not resolve all dependencies for configuration ':classpath'.
   > Could not find org.jetbrains.kotlin:kotlin-gradle-plugin:1.1.0-rc-91.
     Searched in the following locations:
         https://plugins.gradle.org/m2/org/jetbrains/kotlin/kotlin-gradle-plugin/1.1.0-rc-91/kotlin-gradle-plugin-1.1.0-rc-91.pom
         https://plugins.gradle.org/m2/org/jetbrains/kotlin/kotlin-gradle-plugin/1.1.0-rc-91/kotlin-gradle-plugin-1.1.0-rc-91.jar
     Required by:
         project : > gradle.plugin.org.jlleitschuh.gradle:ktlint-gradle:1.0.2
```
The problem goes away if you add this to the top of `build.gradle`:
```groovy
buildscript {
    repositories {
        maven { url 'http://dl.bintray.com/kotlin/kotlin-eap-1.1' }
    }
}
```
I'm a bit puzzled as to the cause since your `1.0.2` plugin appears to depend on kotlin `1.1.1` 🤔.

PR #7, opened 2017-04-23
Simplify kotlin dependencies

jeremymailen:
- Simplify the plugins used to build the plugin since they won't affect the plugin user
- Make plugin dependencies compileOnly since they need not become transitive dependencies for the plugin user
- switch to standard gradle 3.5 wrapper (which supports kotlin build scripts) to avoid kotlin version conflict

I believe it fixes https://github.com/JLLeitschuh/ktlint-gradle/issues/6
At least I was able to test it successfully locally against a kotlin 1.1.1 project with a groovy build script.


build.gradle.kts
```diff
@@ -1,40 +1,23 @@
...
-        maven { setUrl("https://plugins.gradle.org/m2/") }
-        mavenCentral()
-    }
-    dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
-        classpath("com.gradle.publish:plugin-publish-plugin:0.9.7")
-    }
+plugins {
+    id("org.jetbrains.kotlin.jvm").version("1.1.1")
+    id("com.gradle.plugin-publish").version("0.9.7")
```
JLLeitschuh: This actually works as an infix:
```kotlin
id("com.gradle.plugin-publish") version "0.9.7"
```
but this is also fine as it is.

build.gradle.kts
```diff
@@ -1,40 +1,23 @@
...
     mavenCentral()
 }
 
 dependencies {
-    compile(gradleApi())
-    compile(gradleScriptKotlinApi())
-    compile(kotlinModule("gradle-plugin"))
+    compileOnly(gradleApi())
+    compileOnly(gradleScriptKotlinApi())
+    compileOnly(kotlinModule("gradle-plugin"))
```
JLLeitschuh: Can you explain why these are not needed as transitive dependencies?

I honestly don't know.
Are you sure this will work correctly?

jeremymailen: I believe the user of the plugin will need to have a version of gradle which supports build script kotlin and also have the JetBrains kotlin plugin in order to use your plugin anyway. Keeping your specific dependency from becoming a transitive dependency is probably more flexible.

I'm not an expert on publishing gradle plugins, but it appeared what was happening was you had an older version of gradle wrapper and since it was used to build the published plugin it brought an older version of kotlin with it as a transitive dependency.

I can do some additional testing with other user project configurations to make sure.

JLLeitschuh: This is a good explanation. Thanks

JLLeitschuh:
Is this PR in a state where you are happy with me merging it? Is there anything else you would like to add?

jeremymailen:
Yep, looks good. I tested the plugin published to mavenLocal against a few different consuming project types (groovy and kotlin build scripts) and it worked in all cases.

I surprised to find it even worked against projects using the `com.zoltu.kotlin` `1.1.0` plugin instead of the new official JetBrains one and also projects using gradle 3.4.

Looking at the published pom generated before using `gradlew`:

and after:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.jlleitschuh.gradle</groupId>
  <artifactId>ktlint-gradle</artifactId>
  <version>1.0.2</version>
  <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-gradle-plugin</artifactId>
      <version>1.1.0-rc-91</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
</project>
```
seems to indicate the desired effect.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.jlleitschuh.gradle</groupId>
  <artifactId>ktlint-gradle</artifactId>
  <version>1.0.3</version>
</project>
```
Thanks for maintaining this plugin!

PR #10, opened 2017-05-30
Support Gradle 4.0

### LINKED ISSUE #9: Does not work with Gradle version 4
`org/gradle/script/lang/kotlin/KotlinDependencyHandler` has been renamed to `DependencyHandlerScope` which means that this plugin breaks.

Moving forward it may make sense to write an adapter or use the java API which may be a bit more stable across releases.

JLLeitschuh:
This removes the dependency upon the gradle script kotlin API
The API has changed in gradle 4.0 and does not have binary compatibility.

Closes #9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+
+## [Unreleased]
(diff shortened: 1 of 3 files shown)


sureshg:
Thanks for fixing this. When can we expect this to be published on Gradle plugin repo?

JLLeitschuh:
In an ideal world I'd have someone to review these changes, but in a project managed by one person...
I should merge and push a release.

JLLeitschuh:
Fixed in release 2.0.1

PR #11, opened 2017-06-07
Move ktlintCheck tasks to more appropriate verification group.

Tapchicoma:
Actually ktlintCheck tasks do nothing related to code formatting, but
they verify validity of kotlin files.


JLLeitschuh:
You are absolutely correct. Thanks!

Issue #8, opened 2017-05-29
This plugin does not seem to work with Android projects


vRallev:
According to the source and my own tests

JLLeitschuh:
That may be true.

I'm not sure how android projects are structured. I basically look for every source set that the `JavaPluginConvention` is aware of, iterate over each and apply this plugin.

https://github.com/JLLeitschuh/ktlint-gradle/blob/master/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt#L43-L78

What problem are you having??

Note, this plugin does **_not_** automatically configure itself so that it's attached to the `check` task. You have to do this manually.

vRallev:
Android is using a different plugin `kotlin-lint`. A few other things would need to be changed as well, e.g. the source set detection.

JLLeitschuh:
I haven't heard of `kotlin-lint` (link??) is this something you are interested in opening a PR to fix?

I don't work in the android ecosystem and I don't really need the feature. If you are interested in seeing it added please feel free to open a PR.

vRallev:
I have no idea why I wrote kotlin-lint. I actually meant `kotlin-android`. Sorry

Tapchicoma:
So I've added ktlint to my project and see that plugin created tasks for android modules that use kotlin as well, but performs checks only on kotlin files in java module.

Tapchicoma:
I think the problem is that on Android java sourcesets are used for kotlin files as well. So this lines:

```kotlin
val kotlinSourceSet: SourceDirectorySet = (it as HasConvention)
        .convention
        .getPluginHelper<KotlinSourceSet>()
        .kotlin
```
doesn't returns any source dirs

JLLeitschuh:
Pull requests that fix this problem will be accepted.

JLLeitschuh:
Closed by #12

PR #12, opened 2017-06-12
Issue #8 - Add support for Android projects

Tapchicoma:
Fix for issue #8: This pull request adds support for Android projects with applied `kotlin-android` plugin. Also adds applying plugin only to root project - it will scan subprojects and add tasks accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,10 @@ ktlint {
 
 This plugin adds two tasks to every source set: `ktlint[source set name]Check` and `ktlint[source set name]Format`.
 Additionally, a simple `ktlintCheck` task has also been added that checks all of the source sets for that project.
-Similarly, a `ktlintFormat` task has been added that formats all of the source sets. Also same meta tasks are added
-to the root project for convenience - they will call related tasks in subprojects. 
+Similarly, a `ktlintFormat` task has been added that formats all of the source sets.
+
+If project has subprojects - plugin also adds two meta tasks `ktlintCheck` and `ktlintFormat` to the root project that 
+triggers related tasks in subprojects.
 
 
 ## Developers
--- a/README.md
+++ b/README.md
@@ -61,7 +61,8 @@ ktlint {
 
 This plugin adds two tasks to every source set: `ktlint[source set name]Check` and `ktlint[source set name]Format`.
 Additionally, a simple `ktlintCheck` task has also been added that checks all of the source sets for that project.
-Similarly, a `ktlintFormat` task has been added that formats all of the source sets.
+Similarly, a `ktlintFormat` task has been added that formats all of the source sets. Also same meta tasks are added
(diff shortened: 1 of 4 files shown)


JLLeitschuh:
I'd still like to be able to run `:[someProject]:ktlintCheck` to only validate one project.
I work on projects with a lot of subprojects and there are sometimes cases where you don't care about other projects during your development cycle.

JLLeitschuh:
Thanks for the PR! I'll look over it a bit more in a bit when I have a free moment.

Tapchicoma:
You can still apply plugin only for subproject. Applying to root project is optional.

JLLeitschuh:
Also, would you be so kind as to update the change log with your change listed under the header.
```
## [Unreleased]
```

I will update the header when I do a release.

src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -1,8 +1,11 @@
 package org.jlleitschuh.gradle.ktlint
 
+import com.android.build.gradle.BaseExtension
```
Tapchicoma: Checked on pure Kotlin project (without any android dependencies) - works ok. When I remove this `compileOnly` dependency - plugin isn't compiled.

README.md
```diff
@@ -61,7 +61,8 @@ ktlint {
 
 This plugin adds two tasks to every source set: `ktlint[source set name]Check` and `ktlint[source set name]Format`.
 Additionally, a simple `ktlintCheck` task has also been added that checks all of the source sets for that project.
-Similarly, a `ktlintFormat` task has been added that formats all of the source sets.
+Similarly, a `ktlintFormat` task has been added that formats all of the source sets. Also same meta tasks are added
+to the root project for convenience - they will call related tasks in subprojects.
```
Tapchicoma: True that user can see tasks from subprojects as well.

Previously when user applied plugin to the root project that contains subprojects - no ktlint tasks are added, but now user can do that - plugin will add tasks for subprojects and add them as a dependency in the root project meta task.

Will make description more understandable.

JLLeitschuh: Thanks! This is the first plugin I've published so it's very likely I did something goofy.

src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
JLLeitschuh: Okay, thanks for checking!

README.md
Tapchicoma: Fixed in 32db797

Tapchicoma:
Updated CHANGELOG in 704ab7b

Issue #13, opened 2017-07-26
Choose Reporter for Ktlint


lswith:
I'd like to use the checkstyle reporter so that I can easily parse any failures that occur.

JLLeitschuh:
Pull requests are welcome! I don't personally have any use for this feature but if you would like to add it I'd be happy to merge it in and ship a release with it.

PR #16, opened 2017-08-03
Fix plugin uses always default ktlint version.

Tapchicoma:

During working on issue #13 I found out that plugin always uses default extension version of ktlint, no matter what version is defined in users project.


JLLeitschuh:
Looks good. Thanks for the fix.

Issue #25, opened 2017-10-04
Standard output is not closed


hmiyado:
https://github.com/JLLeitschuh/ktlint-gradle/blob/6b6fca879980534c3de94c37a259c205fc44d235/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt#L26

Standard output is opened at this code but this stream is not closed.

JLLeitschuh:
Is it supposed to be? Doesn't Gradle do this for you?

hmiyado:
I cannot delete ktlint.xml because of `java.io.IOException`.
I tried a little research, and found ktlint sample code(below).
In it, standardOutput is closed explicitly.

```kotlin
task ktlintGenerateCheckstyleXML(type: Exec) {
  commandLine 'java', '-cp', configurations.ktlint.join(System.getProperty('path.separator')),
    'com.github.shyiko.ktlint.Main', '--reporter=checkstyle', 'src/**/*.kt'
  standardOutput = new FileOutputStream(new File(buildDir, "ktlint-checkstyle-report.xml"))
  ignoreExitValue = true
  doLast {
    standardOutput.close()
    if (execResult.exitValue != 0) {
      throw new GradleException("ktlint finished with non-zero exit value ${execResult.exitValue}")
    }
  }
}
```

Tapchicoma:
@hmiyado you are right, need to add explicit close.

I will fix it.

PR #27, opened 2017-10-06
Fix report output is always opened since task is created.

hmiyado:
I reported the problem that I cannot delete ktlint output file for `java.io.IOException` at #25 .
I thought the cause was that output stream was not closed.
However after fixed it, I still couldn't delete the file.

It seems that output stream is always opened.
I googled and found answers like [this](https://discuss.gradle.org/t/exec-standardoutput-to-file-in-builddir-but-no-builddir-yet/17913/2) or [this](https://stackoverflow.com/questions/13172215/gradle-output-file-not-up-to-date-if-parsed-from-standardoutput).
According to those, FileOutputStream should open in `doFirst` closure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unreleased]
-(Nothing yet)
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
Yea, this makes sense. Good find, good catch.

JLLeitschuh:
A release is now available with your fix `2.2.1`.

PR #31, opened 2017-10-25
Fix wrong parsing of ktlint version.

### LINKED ISSUE #28: Unable to specify any reporter if ktlint version is bigger than "0.9.x"
There seems to be a bug that prevents you from using a custom reporter when ktlint version is bigger than "0.9.x", eg. "0.10.1".

This is due to comparing strings instead of integers:

```
private fun isReportAvailable(version: String, availableSinceVersion: String): Boolean {
    val versionsNumbers = version.split('.')
    val reporterVersionNumbers = availableSinceVersion.split('.')
    return versionsNumbers[0] >= reporterVersionNumbers[0] &&
            versionsNumbers[1] >= reporterVersionNumbers[1] &&
            versionsNumbers[2] >= reporterVersionNumbers[2]
}
```
To reproduce the issue you can use this configuration:
```
ktlint {
    version = "0.10.1"
    debug = true
    verbose = false
    reporter = "json"
    ignoreFailures = true
}
```
The output is:
```
> Task :app:ktlintItalyCheck
[DEBUG] Discovered ruleset "standard"
[DEBUG] Discovered reporter "plain"
[DEBUG] Discovered reporter "json"
[DEBUG] Discovered reporter "checkstyle"
[DEBUG] Initializing "plain" reporter with {verbose=false}
[DEBUG] {} loaded from .editorconfig
[...]
```

Tapchicoma:
This prevents to use plugin with >0.10.x versions of ktlint.

Fixes #28 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unreleased]
+### Fixed
+- Fixed plugin doesn't apply custom reporter for ktlint versions >0.10.x (#28)
 
 ## [2.2.1] - 2017-10-06
 ### Fixed
--- a/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
+++ b/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
@@ -37,8 +37,8 @@ fun JavaExec.applyReporter(target: Project, extension: KtlintExtension) {
 }
 
 private fun isReportAvailable(version: String, availableSinceVersion: String): Boolean {
-    val versionsNumbers = version.split('.')
-    val reporterVersionNumbers = availableSinceVersion.split('.')
+    val versionsNumbers = version.split('.').map { it.toInt() }
+    val reporterVersionNumbers = availableSinceVersion.split('.').map { it.toInt() }
     return versionsNumbers[0] >= reporterVersionNumbers[0] &&
             versionsNumbers[1] >= reporterVersionNumbers[1] &&
             versionsNumbers[2] >= reporterVersionNumbers[2]


src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
```diff
@@ -37,8 +37,8 @@ fun JavaExec.applyReporter(target: Project, extension: KtlintExtension) {
 }
 
 private fun isReportAvailable(version: String, availableSinceVersion: String): Boolean {
-    val versionsNumbers = version.split('.')
-    val reporterVersionNumbers = availableSinceVersion.split('.')
+    val versionsNumbers = version.split('.').map { it.toInt() }
+    val reporterVersionNumbers = availableSinceVersion.split('.').map { it.toInt() }
```
JLLeitschuh: @shyiko Just want to check with you, do you think you will ever have a release that has some sort of `RC1` specifier on the end?
If not this should be fine.

Tapchicoma: @JLLeitschuh I may apply your proposal to use symver plugin in a separate PR

shyiko: @JLLeitschuh never (and you can count on it). ktlint adheres to strict semver without pre-release|s or build meta.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [unreleased]
+###
```
JLLeitschuh: Can you make this header `Fixed` like the one below.

Tapchicoma: Sorry, missed this. Fixed in updated PR.

JLLeitschuh:
Thanks!

PR #22, opened 2017-09-04
define a different output file for each sourceSet

lampietti:
While the ktlint checkstyle reporter stream Strings in output file, it generates a corrupted XML after the first sourceSet run.

Fix-it by generating one file by sourceSet, with the following naming: `ktlint-${sourceSetName}.${extension.reporter.fileExtension}`

PR #35, opened 2017-10-27
Different output file for every task (fix for #33)

### LINKED ISSUE #33: Same output file for every task
Currently the output of the all the ktlint*Check task is the same XML file:
image
This means that is not possible to have multiple tasks checking different variants in the same gradle execution (this is really common in CI).

My suggestion is to align the output to the one that Lint is using and generate one file per task:
image

This way is possible to generate multiple reports with a single gradle execution.

I have already fixed this issue locally, if you want I can open a Pull Request.

leinardi:
Previously every task was using the same name as output file, so that if
you execute multiple tasks during the same build, every task was
overwriting the previous output.

This commit fix this issue providing a different name per each task. The
name is generated using the same source set name that is used for the
task name. This is also aligned to the output of Android Lint.

This fixes #33

### DIFF
--- a/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
+++ b/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
@@ -21,11 +21,11 @@ enum class ReporterType(val reporterName: String, val availableSinceVersion: Str
 /**
  * Apply reporter to the task.
  */
-fun JavaExec.applyReporter(target: Project, extension: KtlintExtension) {
+fun JavaExec.applyReporter(target: Project, extension: KtlintExtension, sourceSetName: String) {
     if (isReportAvailable(extension.version, extension.reporter.availableSinceVersion)) {
         var reportOutput: FileOutputStream? = null
         doFirst {
-            reportOutput = createReportOutputDir(target, extension).outputStream().also {
+            reportOutput = createReportOutputDir(target, extension, sourceSetName).outputStream().also {
                 this.args("--reporter=${extension.reporter.reporterName}")
                 this.standardOutput = it
             }
@@ -44,8 +44,8 @@ private fun isReportAvailable(version: String, availableSinceVersion: String): B
             versionsNumbers[2] >= reporterVersionNumbers[2]
 }
 
-private fun createReportOutputDir(target: Project, extension: KtlintExtension): File {
-    val reportsDir = File(target.buildDir, "reports/ktlint")
+private fun createReportOutputDir(target: Project, extension: KtlintExtension, sourceSetName: String): File {
+    val reportsDir = File(target.buildDir, "reports")
     GFileUtils.mkdirs(reportsDir)
-    return File(reportsDir, "ktlint.${extension.reporter.fileExtension}")
+    return File(reportsDir, "ktlint-results-$sourceSetName.${extension.reporter.fileExtension}")
 }
--- a/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -144,7 +144,7 @@ open class KtlintPlugin : Plugin<Project> {
             args(runArgs)
         }.apply {
             this.isIgnoreExitValue = extension.ignoreFailures
-            this.applyReporter(target, extension)
+            this.applyReporter(target, extension, sourceSetName)
         }
     }
 


JLLeitschuh:
This is beginning to reach the point where it makes sense to create our own gradle task that allows users to do something like 
```kotlin
tasks.withType<Ktlint> {
    outputDir = "somethingDifferent"
}
```

Not something I expect you to implement in this PR. Just a thought for myself and @Tapchicoma 

Can you post a screenshot of the output directory structure after this change is applied??

leinardi:
Hi @JLLeitschuh, the screenshots are already available in #33, but I add them also here.

Before this PR:
image

After this PR:
image

nonpayItalyCompatDebug is the name of the variant. If you don't have flavors the variant will be just 'main' an the file name will be ktlint-results-main.xml.

src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
```diff
@@ -44,8 +44,8 @@ private fun isReportAvailable(version: String, availableSinceVersion: String): B
             versionsNumbers[2] >= reporterVersionNumbers[2]
 }
 
-private fun createReportOutputDir(target: Project, extension: KtlintExtension): File {
-    val reportsDir = File(target.buildDir, "reports/ktlint")
+private fun createReportOutputDir(target: Project, extension: KtlintExtension, sourceSetName: String): File {
+    val reportsDir = File(target.buildDir, "reports")
```
Tapchicoma: What is the reason for removing `ktlint` dir?

Tapchicoma:
Can you make this output split optional?

src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
leinardi: To align with the behavior like, as you can see in the screenshot, Android Lint. There is a specific reason why the `ktlint` folder is necessary?

leinardi:
I can try, but is this really a necessary feature? What is the advantage of making the split optional? With this PR if you only have one task you will still only have one file, and if you execute multiple tasks you will be able to see the output of every single file. What is the advantage of overriding the previous output during the same gradle execution? You basically run multiple tasks for nothing since you only get the output from the last task executed.

src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
Tapchicoma: My point is that usually this reports are grabbed by the project CI. It is much easier to just grab all files from subfolder in `reports` dir then grab by some mask that can fail after plugin update, because there is new naming strategy.

Tapchicoma:
Ok, now I see your point. It actually fixes the bug when running not meta `ktlintCheck` task, but several tasks in parallel.

Then, yes, this making this optional not needed.

Tapchicoma:
But then what is the difference between this PR and PR #22? (Except a little different naming)

leinardi:
Oh wow, there is no difference with #22, I was not aware of that PR, I just end up making the exact same fix because I need this for my CI since we run multiple task in parallel.

Why was #22 not merged before?

src/main/kotlin/org/jlleitschuh/gradle/ktlint/Reporter.kt
JLLeitschuh: This is also a breaking change and will require a major version bump on the next release.

JLLeitschuh:
#22 was not merged because I dropped the ball. I can merge it if you think the solution is better.

JLLeitschuh:
Unless there is some reason why this solution is better, I think I'll go with #22

leinardi:
Please go ahead and merge #22, it is fair to give him credit since he made the same changes 1 month eardly.

I would still like to do some little small changes after the merge of #22, like calling the file name `"ktlint-results-$sourceSetName.${extension.reporter.fileExtension}"`. What about the directory `ktlint`? Do you still believe is necessary? Usually in CI configurations is common to use regex to select specific files and not get everything is inside a folder (at least this is my personal experience with Jenkins).

JLLeitschuh:
Re `ktlint` directory, the gradle standard is to put reports into their own directories.
For example, findbugs put's it's report in `build/reports/findbugs`. I think it makes sense to follow this standard.
However, we should expose a `Ktlint` task type that a user can do
```
tasks.withType<Ktlint> {
    outputDir = file("${project.buildDir}/reports")
}
```
If that's really what they want to do.

Tapchicoma:
I think it is ok to close this PR as #22 was already merged.

JLLeitschuh:
Closing as resolved in another PR. If this is not the case, please open a new PR with the conflicts resolved. Thanks!

PR #43, opened 2017-11-12
Provide output to console

### LINKED ISSUE #38: Provide output to console
`Ktlint` since `0.10.0` allows to provide multiple outputs.

Plugin should enable plain output by default for console, this will allow to see style failures in console builds.

radimvaculik:
* Output to console
* Support multiple reporters for ktlint >0.10.x

Resolves #38.


radimvaculik:
Hello @JLLeitschuh @Tapchicoma, could you do a code review, please?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -29,11 +35,11 @@ open class KtlintExtension {
...
-     * Report output format.
+     * Report output formats.
      *
      * Available values: plain, plain_group_by_file, checkstyle, json.
      *
-     * Default is plain.
+     * Default is empty.
      */
-    var reporter: ReporterType = ReporterType.PLAIN
+    var reporters: List<ReporterType> = mutableListOf()
```
JLLeitschuh: Very much a breaking change. But it makes sense.
Perhaps this should be a `set` instead of a `list`? Unless ordering really matters?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/Reporter.kt
```diff
@@ -0,0 +1,59 @@
...
+import net.swiftzer.semver.SemVer
+import org.jlleitschuh.gradle.ktlint.KtlintExtension
+import java.io.FileOutputStream
+import java.io.OutputStream
+
+/**
+ * Apply reporter to the task.
+ */
+internal fun JavaExec.applyReporters(target: Project, extension: KtlintExtension, sourceSetName: String) {
+    if (SemVer.parse(extension.version).compareTo(SemVer(0, 10, 0)) >= 0) {
```
JLLeitschuh: Documentation why this version check is necessary would be very helpful.

JLLeitschuh: In Koltin you should be able to do something like `SemVer.parse(extension.version) >= SemVer(0, 10, 0)` so you don't need to have the extra `>= 0`.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -29,11 +35,11 @@ open class KtlintExtension {
...
     /**
-     * Report output format.
+     * Report output formats.
      *
      * Available values: plain, plain_group_by_file, checkstyle, json.
      *
-     * Default is plain.
+     * Default is empty.
      */
-    var reporter: ReporterType = ReporterType.PLAIN
```
JLLeitschuh: Do we want to deprecate this before outright removing it? @Tapchicoma thoughts?

README.md
```diff
@@ -55,7 +55,8 @@ ktlint {
     debug = true
     verbose = true
     android = false
-    reporter = "checkstyle"
+    outputToConsole = true
+    reporters = ["plain", "checkstyle"]
```
JLLeitschuh: This actually isn't valid kotlin. I don't know if groovy will do this coercion either. You may need the actual type `ReporterType.PLAIN` here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
JLLeitschuh: This collection should keep `ReporterType.PLAIN` by default.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/Reporter.kt
```diff
@@ -0,0 +1,59 @@
...
+            if (SemVer.parse(extension.version).compareTo(reporter.availableSinceVersion) >= 0) {
+                var reportOutput: FileOutputStream? = null
+                doFirst {
+                    reportOutput = createReportOutputFile(target, reporter.fileExtension, sourceSetName).outputStream().also {
+                        this.args("--reporter=${reporter.reporterName}")
+                        if (extension.outputToConsole) {
+                            this.standardOutput = object : OutputStream() {
+                                override fun write(b: Int) {
+                                    it.write(b)
+                                    System.out.write(b)
```
JLLeitschuh: Was this left in intentionally??

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: Yeah, I think deprecation will be better. Less painful to update the plugin.

@JLLeitschuh do you know if https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/index.html will work here? Would be very useful to be able to update this field using IDE action.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -14,17 +14,20 @@ import org.gradle.api.tasks.JavaExec
...
-const val CHECK_PARENT_TASK_NAME = "ktlintCheck"
-const val FORMAT_PARENT_TASK_NAME = "ktlintFormat"
+import org.jlleitschuh.gradle.ktlint.reporter.applyReporters
 
 /**
  * Task that provides a wrapper over the `ktlint` project.
  */
 open class KtlintPlugin : Plugin<Project> {
 
+    companion object {
```
Tapchicoma: There is no point to move this constants into `companion object`.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
JLLeitschuh: Probably would. If were were to hard code the refactor suggestion to setting the field to be `setOf(ReporterType.PLAIN)`.

Not 100% certain though.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/Reporter.kt
radimvaculik: Yes, if the `outputToConsole` is `true`, put all output to console as well.

README.md
radimvaculik: @JLLeitschuh Any idea, how to do it wit `setOf()` ?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
radimvaculik: I did it as `setOf(ReporterType.PLAIN)`

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/Reporter.kt
JLLeitschuh: @Tapchicoma Should this be printed to a logger or to `System.out`?

README.md
JLLeitschuh: I don't understand the question

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/Reporter.kt
radimvaculik: Ktlint with version 0.10.0 and lower does not support multiple reporters. So if the developer wants to `outputToConsole` and reporting to file in same time, I need to transfer output from ktlint to console and file in one go.

JLLeitschuh: Right, that makes sense.

README.md
radimvaculik: To be honest, I don't know, how to write the syntax for array/set in gradle file for `reporters`. Is there any way, how to do it? To pass multiple reporters? @JLLeitschuh

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: Our idea is to have deprecated `var reporter` field and new `var reporters` and add following annotation for deprecated one:
```
@Deprecated(
    message = "Ktlint introduced multioutput support since 0.11.1 version",
    replaceWith = ReplaceWith("reporters = setOf(x)", "reporter"))
```

CHANGELOG.md
```diff
@@ -15,9 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [2.3.0] - 2017-11-13
 ### Added
- - Add configuration parameter `android` with default value to false
+ - Add configuration parameter `android` with default value to false (#39)
 ### Changed
  - Define a different output file for each sourceSet
+ - Output to console (#38)
+ - Support multiple reporters for ktlint >0.10.x (#38)
```
JLLeitschuh: This shouldn't be listed under this release. `2.3.0` has already been released.

Tapchicoma:
If I apply following to [build.gradle.kts](samples/kotlin/build.gradle.kts):

```
configure<KtlintExtension> {
    reporters = setOf(PLAIN, CHECKSTYLE)
}
```
and introduce some style error in [main.kt](samples/kotlin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/sample/kotlin/main.kt), I see error in the console, but no reports in build/reports are created.

radimvaculik:
@Tapchicoma You need to specify the version of ktlint to be higher than 0.10.x.

Tapchicoma:
@radimvaculik my bad, totally forget to use higher ktlint version. Checked again - it works 👍

radimvaculik:
@Tapchicoma @JLLeitschuh Ready to merge for me. Could you do a code review, please?
I changed `Set<ReporterType>` to `Array<ReporterType>` as `Array` is only collection supported by gradle plugin. I also lowercased `ReporterType` enums, because it's case-sensitive, while matching string with enum.

radimvaculik:
@Tapchicoma @JLLeitschuh Could you do a code review, please?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -29,11 +35,14 @@ open class KtlintExtension {
...
+    var reporters: Array<ReporterType> = arrayOf(ReporterType.plain)
+
+    @Deprecated(message = "Ktlint introduced multi output support since 0.11.1 version", replaceWith = ReplaceWith("reporters = arrayOf(x)", "reporter"))
```
Tapchicoma: Seems `reporters = arrayOf(x)` not working as expected - it produces `reporters = arrayOf(x)`.

Change it to `ReplaceWith("reporters = arrayOf()")`

Tapchicoma:
@radimvaculik Overall looks ok. I've tested on several projects - works!

After this small changes fix I think it is ready to be merged.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReporterType.kt
```diff
@@ -0,0 +1,10 @@
+package org.jlleitschuh.gradle.ktlint.reporter
+
+import net.swiftzer.semver.SemVer
+
+enum class ReporterType(val reporterName: String, val availableSinceVersion: SemVer, val fileExtension: String) {
+    plain("plain", SemVer(0, 9, 0), "txt"),
```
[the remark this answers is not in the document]
JLLeitschuh: I agree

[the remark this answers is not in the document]
radimvaculik: But then it won't work with Groovy parsing for:
```
reporters = ["plain", "checkstyle"]
```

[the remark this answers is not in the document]
JLLeitschuh: How does groovy figure out which enum to map to? I'm sure there's a method you can implement where you could do something like `equalsIgnoreCase` and it would work fine.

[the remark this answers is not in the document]
radimvaculik: @JLLeitschuh this is my first contribution to gradle plugin (and it's written in kotlin), so I have no idea, how to do it. I spent couple hours debugging, how it works etc.

Tapchicoma:
@radimvaculik can you grant permission to update your PR? I want to finish it.

Tapchicoma:
Just for History link to Gradle issue: https://github.com/gradle/gradle/issues/4030

Tapchicoma:
@JLLeitschuh can you check again this PR?

radimvaculik:
@Tapchicoma Looks good to me :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -28,12 +35,25 @@ open class KtlintExtension {
...
      *
      * Available values: plain, plain_group_by_file, checkstyle, json.
      *
-     * Default is plain.
+     * Default is none.
      */
-    var reporter: ReporterType = ReporterType.PLAIN
+    @Deprecated(message = "Ktlint introduced multi output support since 0.11.1 version",
+            replaceWith = ReplaceWith("reporters = arrayOf()", "reporter"))
+    var reporter: DeprecatedReporterType? = null
```
JLLeitschuh: Kind of a big change to change a default don't you think?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/ReporterType.kt
```diff
@@ -0,0 +1,11 @@
+package org.jlleitschuh.gradle.ktlint
+
+import net.swiftzer.semver.SemVer
+
+@Deprecated("Moved to org.jlleitschuh.gradle.ktlint.reporter package")
```
JLLeitschuh: :x: For the next few releases use a typealias, then we can move it.
See the section here about "managing change"
https://opencredo.com/some-uses-for-type-aliases-in-kotlin/

I think if you leave the typealias behind and move the class then everything should still compile for old users.
Not 100% certain. It would nice if you could test that to make sure I'm right.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -1,13 +1,16 @@
...
+
 /**
  * Extension class for configuring the [KtlintPlugin].
  */
 open class KtlintExtension {
     /**
      * The version of ktlint to use.
      */
-    var version = "0.9.2"
+    var version = "0.15.0"
```
JLLeitschuh: 👍

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: Hm. After thinking about it again - you are right. There maybe a users that used default config and assume that `PLAIN` report will be generated.

Tapchicoma: Tough I will leave possibility to unset the value, so plugin will not produce unneeded report type.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/ReporterType.kt
Tapchicoma: Nice trick. Checked and it actually works for users that will update from previous plugin versions.

Tapchicoma:
@JLLeitschuh addressed code review issues

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -1,13 +1,15 @@
 package org.jlleitschuh.gradle.ktlint
 
+typealias ReporterType = org.jlleitschuh.gradle.ktlint.reporter.ReporterType
```
JLLeitschuh: Mark this as `@Depricated` with a mention of what the replacement should be.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -28,12 +34,25 @@ open class KtlintExtension {
...
+    /**
+     * Report output formats.
+     *
+     * Available values: plain, plain_group_by_file, checkstyle, json.
+     *
+     * **Note** for Gradle scripts: for now all values should be uppercase due to bug in Gradle.
+     *
+     * Default is empty.
+     */
+    var reporters: Array<ReporterType> = emptyArray()
```
JLLeitschuh: Shouldn't this default to just `PLAIN`?

JLLeitschuh: Why is this an `Array` instead of a `List`??

samples/kotlin-ks/build.gradle.kts
```diff
@@ -20,5 +20,6 @@ dependencies {
 
 configure<KtlintExtension> {
     verbose = true
-    reporter = ReporterType.JSON
+    outputToConsole = true
+    reporters = arrayOf(ReporterType.CHECKSTYLE, ReporterType.JSON)
```
JLLeitschuh: The extension should support doing `reporters  = listOf(ReporterType.CHECKSTYLE, ReporterType.JSON)`

Issue #58, opened 2018-02-14
Add a way to run check for non-JVM projects


mverleg:
Feature request: _Could some way be added to check code for other platforms than JVM?_

I have a multiplatform module but the kotlin-common and kotlin-javascript code is skipped. Most of the code is kotlin-common, which should be a subset of kotlin-JVM I think.

I'm not sure how hard this is but I'll just suggest it (I'm not really sure how the platform relates to style checks).

mverleg:
(I'm afraid I can't be much help implementing the feature because I'm not great with Gradle, and Gradle hates me).

mverleg:
I found there's another solution I can use. I'll leave the issue for others to evaluate the usefulness (I feel like most other people are on JVM).

JLLeitschuh:
Do you want to share the alternative solution you came up with?

mverleg:
Thanks for the suggestions :-) I spent some time with Gradle stuff but I still get stuck a lot.

I wasn't sure if it was considered impolite to mention 'competing' libraries here... The solution was that I'm using Spotless because of the non-JVM code.

Tapchicoma:
Indeed would be nice to add such support. Should be not difficult - just add `kotlin2js` and other types plugins to supported one.

Actually we can start with js support and and the introduce it for native (version `0.6` introduces Gradle plugin).

@mverleg would be really nice if you can introduce some multiplatform sample for this project.

JLLeitschuh:
> I wasn't sure if it was considered impolite to mention 'competing' libraries here...

I'm one of the maintainers of Spotless and I use it at my company over my own plugin (blasphemous, I know 😆).
You're amongst friends here.

mverleg:
@Tapchicoma Do you mean a sample project to try this on? I happen to have a demo project for multiplatform Kotlin Gradle setup with some subprojects, interdependencies and actual/expect. It has no real code, it's just to experiment with multiplatform setup. Welcome to try with that. https://github.com/mverleg/kotlin_multiplatform_gradle_demo . It doesn't have Kotlin Native yet, but I might add it when it's more officially compatible with kotlin-common.

Tapchicoma:
@mverleg this project has some samples and it would be nice to also add some multiplatform sample, so we can test plugin setup and see that it works correctly.

Tapchicoma:
Missing plugins:
- kotlin2js (javascript)
- konan (native)
- kotlin-platform-common
- kotlin-platform-js
[...]

PR #61, opened 2018-02-17
Add multiplatform support

mverleg:
This is a simple multiplatform project that can be used as a sample, and to work on issue #58.

So currently _the linting does not work_ on common or JS platforms! It does builds and run everything, and does linting for JVM.

I think this can either be left open until #58 is resolved to serve as example, or be merged now to help with testing #58.


Tapchicoma:
Or you can enable _Allow edits from maintainers_, then I can try to update this PR.

mverleg:
I think that setting is on. It's going to be a few days to look at it myself, so very welcome to make changes.

Tapchicoma:
- Moved multiplatform samples to use the same root samples project.
- Convert them to use Kotlin DSL in build scripts.
- Add support for multiplatform kotlin plugin in plugin itself.

Tapchicoma:
@JLLeitschuh can you check this PR? Thanks.

samples/kotlin-multiplatform-js/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
+import org.jlleitschuh.gradle.ktlint.KtlintExtension
+
+apply {
+    plugins.apply("org.jlleitschuh.gradle.ktlint")
```
JLLeitschuh: Uhhh... This doesn't seem right.

samples/kotlin-multiplatform-jvm/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
+import org.jlleitschuh.gradle.ktlint.KtlintExtension
+
+apply {
+    plugins.apply("org.jlleitschuh.gradle.ktlint")
```
JLLeitschuh: This also doesn't seem right.

samples/kotlin-multiplatform-js/build.gradle.kts
Tapchicoma: This is how all samples are implemented, check this line in main `build.gradle.kts`: https://github.com/JLLeitschuh/ktlint-gradle/blob/27b5e6586c1b1973ba19209b913ea53a3ff583cd/build.gradle.kts#L8

Tapchicoma: So each sample module apply plugin from included plugin project using `plugins.apply()`

JLLeitschuh: I think that this is what you want instead.
```kotlin
plugins {
    // This is a better API as the Kotlin DSL can automatically generate accessors for the `KtlintExtension` 
    id("org.jlleitschuh.gradle.ktlint")
}
apply {
    plugin("kotlin-platform-js")
}
```

If you use the above syntax, you should get this for free:
```kotlin
ktlint {
    // Your configuration goes here.
}
```

Let me know if this isn't the case.

Tapchicoma: But will this:

Tapchicoma: Sorry, maybe I was not clear. 

Current approach uses compiled plugin from plugin project included build: https://github.com/JLLeitschuh/ktlint-gradle/blob/27b5e6586c1b1973ba19209b913ea53a3ff583cd/settings.gradle.kts#L40

So when someone changes something in plugin project samples will compile it and apply to their configuration.

By using `plugins {}` block samples will always try to download latest plugin version from some artifactory and I can't find a way to tell it to use included build.

Also I think this part of Gradle documentation relates to this particular use-case: https://docs.gradle.org/4.5.1/userguide/plugins.html#sec:applying_plugins_buildscript
And it advises to use current approach.

JLLeitschuh: Ah, I see. Okay.
This is related to this issue:
https://github.com/gradle/gradle/issues/2528

Okay, fair enough.
Regardless, `plugins` doesn't need to be inside of the `apply` as the receiver of the `plugins` call is on `Project`.
So you can do this instead:
```kotlin
apply {
    plugin("kotlin-platform-js")
}
plugins.apply("org.jlleitschuh.gradle.ktlint")
```

Tapchicoma:
@JLLeitschuh addressed review

JLLeitschuh:
Awesome! Merge at will!

PR #64, opened 2018-03-01
Add kotlin native support

Tapchicoma:
#58: Add support for kotlin native (konan) plugin.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,6 +2,7 @@ buildscript {
     repositories {
         jcenter()
         google()
+        maven("https://dl.bintray.com/jetbrains/kotlin-native-dependencies")
     }
 
     dependencies {
@@ -11,6 +12,7 @@ buildscript {
 plugins {
     kotlin("jvm") version "1.2.21" apply false
     id("com.android.application") version "3.0.1" apply false
+    id("konan") version "0.6" apply false
 }
 
 allprojects {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Added
  - Support for Kotlin javascript (kotlin2js) (#58)
  - Support for Kotlin multiplatform projects (#58)
+ - Support for Kotlin native (konan) (#58)
 ## Changed
  - ?
 ## Fixed
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -15,12 +15,14 @@ version = "3.0.2-SNAPSHOT"
 repositories {
     jcenter()
     google()
+    maven("https://dl.bintray.com/jetbrains/kotlin-native-dependencies")
 }
 
 dependencies {
     compileOnly(gradleApi())
     compileOnly(kotlin("gradle-plugin", "1.2.21"))
     compileOnly("com.android.tools.build:gradle:3.0.0")
+    compileOnly("org.jetbrains.kotlin:kotlin-native-gradle-plugin:0.6")
     compile("net.swiftzer.semver:semver:1.0.0")
 
     /*
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -5,6 +5,7 @@ import com.android.build.gradle.FeaturePlugin
 import com.android.build.gradle.InstantAppPlugin
 import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.TestPlugin
+import net.swiftzer.semver.SemVer
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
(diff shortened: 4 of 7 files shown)


Tapchicoma:
@JLLeitschuh last of new plugins to support

JLLeitschuh:
What happens if you have both the native plugin applied and the js plugin applied?
This might be a larger issue outside the scope of this PR but if it is an issue it should be fixed before release.

Tapchicoma:
Thoretically plugin will create two sets of tasks for both native and js with common meta tasks: `ktlintCheck` and `ktlintFormat`. Will check this scenario tomorrow.

Tapchicoma:
@JLLeitschuh after adding `kotlin2js` plugin to native sample here is output of running `ktlintCheck`:
```text
./gradlew -m :samples:kotlin-native:ktlintCheck
:samples:kotlin-native:ktlintKotlin-nativeCheck SKIPPED
:samples:kotlin-native:ktlintMainCheck SKIPPED
:samples:kotlin-native:ktlintTestCheck SKIPPED
:samples:kotlin-native:ktlintCheck SKIPPED
```

So my guess was right.

JLLeitschuh:
Hrm, seems like redundant execution. But it won't create a problem I suppose.

build.gradle.kts
```diff
@@ -11,6 +12,7 @@ buildscript {
 plugins {
     kotlin("jvm") version "1.2.21" apply false
     id("com.android.application") version "3.0.1" apply false
+    id("konan") version "0.6" apply false
```
JLLeitschuh: Why is this necessary in the base plugin. What happens if the plugin isn't added to the entire build's classpath and only to the one subproject?

Tapchicoma: According to Gradle docs:

Tapchicoma: According to Gradle docs:

> Applying plugins to subprojects

> If you have a multi-project build, you probably want to apply plugins to some or all of the subprojects in your build, but not to the root or master project. The default behavior of the plugins {} block is to immediately resolve and apply the plugins. But, you can use the apply false syntax to tell Gradle not to apply the plugin to the current project and then use apply plugin: «plugin id» in the subprojects block:

I also tried it and I see following error:
```text
Caused by: java.lang.NoClassDefFoundError: org/jetbrains/kotlin/gradle/plugin/KonanExtension
```

Tapchicoma: My guess that due to ktlint uses root project classpath - it doesn't see module additional classpath with kotlin-native plugin classes.

JLLeitschuh: We shouldn't require users to apply the `konan` plugin to use this plugin.

Tapchicoma: `konan` plugin has `compileOnly` dependency, so it is totally optional whether it is applied or not. Projects that doesn't apply `konan` plugin will not be affected.

Tapchicoma: Regarding this `NoClassDefFoundError` I think if plugin is applied in subproject, Gradle creates the classloader that is only limited to this subproject and is not available to the parent scope (?). Check this classes:
[DefaultInjectedPluginDependencies](https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/initialization/definition/DefaultInjectedPluginDependencies.java)
[ClassLoaderScope](https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/api/internal/initialization/ClassLoaderScope.java)
[ClassPathPluginResolution](https://github.com/gradle/gradle/blob/master/subprojects/plugin-use/src/main/java/org/gradle/plugin/use/resolve/internal/ClassPathPluginResolution.java)

So Ktlint-gradle sees that plugin is applied, can get extension by String name, but can't cast it to extension class as this class is not in main project classpath.

IMHO it is ok to release new version with such limitation.

JLLeitschuh: It seems that your getting `NoClassDefFoundError` exceptions in our samples prove your interpretation incorrect(?).

When you apply a plugin to the parent, even with `apply: false` it adds it to the classpath for all subprojects.

But when you only apply it to one of the children, siblings of that child won't see the dependent plugin on their classpaths. That's why you are seeing the `NoClassDefFoundError`.

JLLeitschuh: I really don't think this is a gradle problem, I think this is just how the Java classloader behaves.
I'll checkout your branch and try it though.

JLLeitschuh: Okay, I think I understand what you're saying.

Basically, the thing I want to ensure is that if someone has a simple build and they don't have the `konan` plugin applied, they don't get a `NoClassDefFoundError`.

As long as users won't get `NoClassDefFoundError` I'm fine with this.

JLLeitschuh: I think that there is a bug in the Gradle Classloader where:
```
target.pluginManager.withPlugin("konan", /** This get's executed **/)
```
Even though the `konan` classes aren't added to the classloader for the plugin's execution environment. This may be an intractable in the JVM. But I don't know.

@Bamboo or @Oehme, do you have any thoughts?

oehme: I'm sorry, but I don't understand the problem. Can you give me an example and the error it produces?

oehme: There is no bug here - The root project can't see classes from its children (only the other way round).

Generally you should always add all plugins to the classpath in your root project to avoid issues like this.

JLLeitschuh: Thank you so much for your feedback.

> This is working as intended - The root project can't see classes from its children (only the other way round).

Right, this is exactly what we are seeing, and I think this behaviour is correct.
What we are seeing that confuses me is that in a sub projects, we apply the plugin in the supproject like this:
```
plugins {
    id("konan") version "0.6"
}

plugins.apply("org.jlleitschuh.gradle.ktlint")
```
We get an exception when configuring that subproject caused by a:
`java.lang.ClassNotFoundException: org.jetbrains.kotlin.gradle.plugin.KonanExtension`.

I would expect that applying the `konan` plugin in the subproject would add it to the classpath so that when the
```
plugins.apply("org.jlleitschuh.gradle.ktlint")
```
is evaluated, the `konan` plugin would be on the classpath already.

oehme: The ktlint plugin was added to the root project, so it can't see the konan plugin in the subproject.

JLLeitschuh: Oh, wow, that makes so much more sense. Thank you for explaining that.

oehme: Our plan for the future is to have plugins defined centrally and forbid having different classloaders for the subprojects. This should away issues like this one.

Issue #17, opened 2017-08-03
Tests are missing


Tapchicoma:
This plugin needs some tests, so it will be easier to add a new features and check already existing behaviour.

@JLLeitschuh you can assign it to me, will try to add them when I will have spare time.

JLLeitschuh:
I know that Junit5 is using something clever for testing their gradle plugin. We should try to use whatever they are using.

Also, if you can drive the tests with Junit5 that would be really cool but not necessary.

JLLeitschuh:
@Tapchicoma Do you want to take the code from @eskatos's PR and drop it in to the repo?

JLLeitschuh:
Closed by c31eae2

PR #72, opened 2018-03-14
Introduce JUnit & Gradle TestKit based integration test

eskatos:
This PR introduces Gradle TestKit to integration test the plugin.

It does so by using JUnit 4. JUnit 5 could be used as the runtime or the engine but this is not mandatory. Feel free to amend if you feel strong about it. Note that some equivalent of the `TemporaryFolder` rule would be required.

In the same changeset, the `java-gradle-plugin` has been introduced. This removes the need for the plugin descriptor properties and allow TestKit to automatically see the plugin classpath when running integration tests.

See #17


plugin/build.gradle.kts
```diff
@@ -20,16 +21,19 @@ repositories {
 
 dependencies {
     compileOnly(gradleApi())
-    compileOnly(kotlin("gradle-plugin", "1.2.21"))
-    compileOnly("com.android.tools.build:gradle:3.0.0")
-    compileOnly("org.jetbrains.kotlin:kotlin-native-gradle-plugin:0.6")
-    compile("net.swiftzer.semver:semver:1.0.0")
+    implementation(kotlin("gradle-plugin", "1.2.21"))
```
Tapchicoma: Can you change back to `compileOnly`? This 3 dependencies should be optional, so users of this plugin will not get them as a transitive dependency.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -0,0 +1,77 @@
...
+                repositories {
+                    gradlePluginPortal()
+                }
+            """.trimIndent())
+            resolve("src/main/kotlin").mkdirs()
+        }
+    }
+
+    @Test
+    fun `check tasks`() {
```
Tapchicoma: Maybe better to split it into two test cases:
- \`Should fail check on failing sources\`
- \`Should succeed check on clean sources\`

Tapchicoma:
BTW thanks for providing example how to write tests for Gradle plugins!

plugin/build.gradle.kts
```diff
@@ -40,6 +44,15 @@ configure<PublishingExtension> {
     }
 }
 
+gradlePlugin {
+    (plugins) {
```
JLLeitschuh: I hate that you have to do this, but I understand why it's necessary.

plugin/build.gradle.kts
```diff
@@ -20,16 +21,19 @@ repositories {
...
+    implementation("net.swiftzer.semver:semver:1.0.0")
 
     /*
      * Do not depend upon the gradle script kotlin plugin API. IE: gradleScriptKotlinApi()
      * It's currently in flux and has binary breaking changes in gradle 4.0
      * https://github.com/JLLeitschuh/ktlint-gradle/issues/9
      */
+
+    testImplementation(gradleTestKit())
+    testImplementation("junit:junit:4.12")
```
JLLeitschuh: :o: Just curious why you chose Junit 4 over Junit 5. I've been using Junit 5 for all my newest projects.

JLLeitschuh:
Not just that, thanks for providing example how to write tests for Gradle plugins in Kotlin!

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
eskatos: Good point, done

plugin/build.gradle.kts
eskatos: Habits and the requirement of `TemporaryFolder`.
I didn't want to introduce another home grown equivalent for JUnit 5.

plugin/build.gradle.kts
eskatos: Using `compileOnly` here is unfortunately not supported by TestKit automatic plugin classpath injection into tests. It can be done with some additional gymnastic, I'll look into it but it will make the testing setup more complicated.

plugin/build.gradle.kts
JLLeitschuh: Fair enough.

plugin/build.gradle.kts
JLLeitschuh: We don't want to add more dependencies than are strictly necessary to run the plugin.

JLLeitschuh:
@eskatos Please rebase on master.

eskatos:
I merged `master` into my branch and amended the PR with dependencies back to `compileOnly`.

plugin/build.gradle.kts
Tapchicoma: Would be nice to introduce JUnit version in `PluginVersions`

PR #74, opened 2018-03-19
Organize and update dependencies

Tapchicoma:
- Update kotlin to `1.2.30` version
- Update Gradle to `4.6` version
- Update default Ktlint version in extension to `0.19.0` version

Add a little dependency organization.


plugin/buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -0,0 +1,9 @@
+object Versions {
+    val kotlin = "1.2.30"
+    val ktlintPlugin = "3.1.0"
+    val gradlePublishPlugin = "0.9.10"
+    val androidPlugin = "3.0.1"
+    val kotlinNativePlugin = "0.6.1"
+    val semver = "1.0.0"
+    val gradleWrapper = "4.6"
+}
```
JLLeitschuh: ❓ Some redundancy here? Any way to reduce it?

Tapchicoma: I tried with `ln -s` as *plugin* is a separate Gradle project, but got `Redeclaration: Versions` error on compile. Can't figure out any other solution.

JLLeitschuh: Use a different name for the two classes? Call one `BuildSrcVersions` and the other something else?

Tapchicoma:
@JLLeitschuh check updated commit. Now samples dependencies depend on plugin dependencies versions.

buildSrc/src/main/kotlin/PluginDependencies.kt
```diff
@@ -0,0 +1 @@
+../../../../plugin/buildSrc/src/main/kotlin/PluginDependencies.kt
```
JLLeitschuh: Is this a symlink or something??

Tapchicoma: yes, symlink. Without it `samples` project doesn't see `PluginVersions` class.

PR #84, opened 2018-04-18
Use relative for input file path sensitivity

wolfs:
This allows enabling caching for the check task. Without adding a path
sensitivity, caching the check task does not make much sense, since it
would require using the same absolute path to the source files for all
workspaces which want to get cache hits.

See https://guides.gradle.org/using-build-cache/#relocatability.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -17,6 +17,7 @@ import org.gradle.api.plugins.AppliedPlugin
 import org.gradle.api.plugins.Convention
 import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.tasks.JavaExec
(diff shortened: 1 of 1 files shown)


wolfs:
I also tried adding output files to the check task, but this turned out to be too complicated as long as the task is a `JavaExec` task and I can use the runtime API for declaring inputs and outputs only.

I would suggest creating a [custom task type](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:task_input_output_annotations) which declares its inputs and outputs via annotations and calls [`javaexec`](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:javaexec(org.gradle.api.Action)) in its task action. Then it should be easy to declare the outputs. I don't know if I will find time to prepare a PR for that.

If someone else has time to give it a shot I would be happy to review it.

JLLeitschuh:
Yea, something that I've had bouncing around my head (but haven't had time to actually implement or do was to make the `Ktlint` task a `SourceTask`. #85

JLLeitschuh:
@Tapchicoma If this looks good to you feel free to merge.

Issue #85, opened 2018-04-18
Make the Ktlint task a SourceTask


JLLeitschuh:
Instead of a `JavaExec` task, the `Ktlint` task should be [SourceTask](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceTask.html).

Tapchicoma:
@JLLeitschuh is it still actual after #90 was merged?

PR #90, opened 2018-04-26
Add task class for KtlintCheck

wolfs:
This PR adds an actual task class for the ktlint check task. This allows declaring outputs for configured reports.

The task class does not yet support the old report command line arguments (single report only).
Moreover, it currently works only on Gradle 4.7, since it uses [`@Nested` on a `Map`](https://docs.gradle.org/4.7/release-notes.html#support-for-mapped-nested-inputs) for declaring the outputs of the report.

If the plugin maintainers think this is a good direction, then I can address the remaining problems like supporting older ktlint versions or older Gradle versions.

This PR addresses part of #85. I ended up not using source task since it does not provide any benefits in this simple case.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -0,0 +1,75 @@
...
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Nested
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+import org.jlleitschuh.gradle.ktlint.reporter.KtlintReport
+import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
+
+open class KtlintCheck : DefaultTask() {
```
eskatos: :o: Implementing `VerificationTask` would make sense

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -0,0 +1,75 @@
...
+    @get:Internal
+    val sourceDirectories: ConfigurableFileCollection = project.files()
+
+    @get:SkipWhenEmpty
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    @get:InputFiles
+    val sources: FileTree = sourceDirectories.asFileTree.matching { it.include("**/*.kt") }
+
+    @get:Input
+    val verbose: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType)
```
eskatos: :x: `Boolean::class.java` should be enough

There are many other instances of this below, not commented on each one

wolfs: If I use `Boolean::class.java` I get:
```
Cannot get the value of a property of type boolean as the provider associated with this property returned a value of type java.lang.Boolean.
```
when running the tests.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
wolfs: Implementing `VerificationTask` would somewhat force me to drop the `Property<Boolean>` type for `ignoreFailures`. Then I would need to use convention mapping, right?

JLLeitschuh: What about `SourceTask`?

eskatos: About `SourceTask`, the extra use case of including/excluding individual files could be tackled later, as a separate PR.

About `VerificationTask`, that's unfortunate but I see your point and agree @wolfs

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
eskatos: Ah right, fair enough

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -210,27 +209,36 @@ open class KtlintPlugin : Plugin<Project> {
...
-            this.isIgnoreExitValue = extension.ignoreFailures
-            this.applyReporters(target, extension, sourceSetName)
+            classpath.setFrom(ktLintConfig)
+            sourceDirectories.setFrom(kotlinSourceDirectories)
+            verbose.set(target.provider { extension.verbose })
+            debug.set(target.provider { extension.debug })
+            android.set(target.provider { extension.android && SemVer.parse(extension.version) >= SemVer(0, 12, 0) })
+            ignoreFailures.set(target.provider { extension.ignoreFailures })
+            outputToConsole.set(target.provider { extension.outputToConsole
+            })
```
JLLeitschuh: Weird formatting here.

wolfs:
@JLLeitschuh Thank you for the review! I addressed/answered your comments.

How are the requirements regarding support for older Gradle/ktlint versions? What do I still need to address?

JLLeitschuh:
I'd feel more comfortable merging this PR after we've come to some sort of conclusion WRT the discussion here: https://github.com/gradle/kotlin-dsl/issues/597

I just want to make sure that we are really committing to using the best practices regarding API we are exposing.

JLLeitschuh:
> How are the requirements regarding support for older Gradle/ktlint versions?

As I'm not currently even using this plugin myself, I'm going to defer to @Tapchicoma about this.

This change doesn't seem to be an API breaking change so I'm fine with a minor version bump on this one.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginVersionTest.kt
```diff
@@ -38,6 +40,8 @@ class KtlintPluginVersionTest : AbstractPluginTest() {
 
     @Test
     fun `with ktlint version less than 0_20`() {
+        withCleanSources()
```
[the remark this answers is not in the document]
JLLeitschuh: They are still here.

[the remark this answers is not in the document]
wolfs: They should be gone now.

Tapchicoma:
> How are the requirements regarding support for older Gradle/KtLint versions?

I am thinking that at some point plugin should drop support for older versions of KtLint and Gradle, to simplify code/reduce effort of supporting it. We can do it in this PR as well.

@wolfs what is desired minimal supported versions of Gradle and KtLint in this PR?

> This change doesn't seem to be an API breaking change so I'm fine with a minor version bump on this one.

Indeed it is not changing the API, but on the other hand will break update for those who is still using old Gradle wrapper/set older KtLint version in extension (I wonder if there such installations :thinking:), so, IMHO, better major version bump.

Tapchicoma:
@wolfs I see that KtLint mininal version is `0.10.0` that introduced multiple reporters - it is ok. As for Gradle 4.7 - it is questionable - people not updating Gradle wrapper so fast, so better to support older versions. Ideally starting 4.0.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -0,0 +1,106 @@
...
+                        throw GradleException(
+                                "The output path passed `$reportOutputPath` contains spaces. Ktlint versions less than 0.20.0 do not support this.")
+                    }
+                }
+            }
+        }
+        project.javaexec {
+            it.classpath = classpath
+            it.main = "com.github.shyiko.ktlint.Main"
+            it.args(sourceDirectories.map { "${it.path}/**/*.kt" })
```
JLLeitschuh: Include `*.kts` incase someone is linting a `buildSrc` project with script based plugins.

wolfs:
@Tapchicoma The PR is now compatible with Gradle 4.3. Being compatible with Gradle 4.0 is somewhat hard, since the methods/types for `Property`s changed between Gradle 4.2 and Gradle 4.3.
Would Gradle 4.3 be OK?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
wolfs: Do `*.kts` files lint well?

Tapchicoma:
@wolfs yes, Gradle 4.3 should be fine :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
Tapchicoma: @wolfs according to https://github.com/shyiko/ktlint/issues/19 since `0.3.0` release

JLLeitschuh:
@wolfs Are getting close to having this PR ready to be merged?

JLLeitschuh:
I'm fine using the `Provider` API for now, yes. Still not a huge fan.

Can you resolve the comment and we can merge this in? Please make a note of your changes in the changelog.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
eskatos: Making that change sounds orthogonal to this PR.
Even though it's a very interesting one :)

wolfs:
@JLLeitschuh Comment resolved (now `**/kts` files are checked, too) and updated changelog.

CHANGELOG.md
```diff
@@ -7,16 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 ## Unreleased [3.4.0-SNAPSHOT]
 ### Added
+ - Also check `*.kts` files in Kotlin source directories
+ - Use a cacheable task for the KtLint check
 ### Changed
   - Update Kotlin to 1.2.41 version
   - Update Gradle wrapper to 4.7 version
 ### Fixed
+### Removed
+  - KtLint versions prior to 0.10.0 are not supported any more
```
JLLeitschuh: Should be: "anymore"

CHANGELOG.md
```diff
@@ -7,16 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 
 ## [3.3.0] - 2018-4-24
 ### Added
  - Check for spaces in output path for KtLint versions earlier
 then 0.20.0 (#83)
- - Use relative for input file path sensivity (#67)
+ - Use relative for input file path sensitivity (#67)
```
JLLeitschuh: 👍

CHANGELOG.md
wolfs: Fixed.

JLLeitschuh:
@wolfs I just want to make sure that the feature added by this PR didn't get dropped in all the changes you made:
https://github.com/JLLeitschuh/ktlint-gradle/pull/89

wolfs:
@JLLeitschuh Good point. I re-added the `ruleSets` feature.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -0,0 +1,128 @@
...
+                        throw GradleException(
+                                "The output path passed `$reportOutputPath` contains spaces. Ktlint versions less than 0.20.0 do not support this.")
+                    }
+                }
+            }
+        }
+        project.javaexec {
+            it.classpath = classpath
+            it.main = "com.github.shyiko.ktlint.Main"
+            it.args(sourceDirectories.flatMap { dir -> KOTLIN_EXTENSIONS.map { extension -> "${dir.path}/**/*.$extension" } })
```
JLLeitschuh: I'm curious why you don't just use the `sources` property you already have in scope for this?

wolfs: Sources are all the Kotlin files in scope. This is possibly a big argument list. If you think Ktlink can happily cope with having all the to be checked `.kt` files as separate arguments on the commandline than that would be fine. I am somewhat concerned about the length of the command line - especially on Windows.

CHANGELOG.md
```diff
@@ -7,16 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 ## Unreleased [3.4.0-SNAPSHOT]
 ### Added
+ - Also check `*.kts` files in Kotlin source directories
+ - Use a cacheable task for the KtLint check
 ### Changed
   - Update Kotlin to 1.2.41 version
   - Update Gradle wrapper to 4.7 version
 ### Fixed
+### Removed
+  - KtLint versions prior to 0.10.0 are not supported anymore
```
Tapchicoma: Please also add that Gradle is supported from version `4.3`

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.TaskAction
+import org.jlleitschuh.gradle.ktlint.reporter.KtlintReport
+import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
+import java.io.File
+import javax.inject.Inject
+
+@CacheableTask
+open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : DefaultTask() {
```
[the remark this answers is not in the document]
wolfs: We love annotations ;)
Is there a less verbose way to do constructor injection in Kotlin?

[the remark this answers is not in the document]
JLLeitschuh: Nope, this is the way to do constructor injection.

I'm actually really pleased that in Kotlin, constructor injection is actually simpler than field injection.

Because of this, it encourages developers to use Constructor injection.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
JLLeitschuh: Totally fair.

Tapchicoma:
@wolfs thanks for implementing this changes!

PR #96, opened 2018-05-27
Add apply to IDEA task.

### LINKED ISSUE #91: Add support for apply-to-idea-project formating command
I added a plugin to the project.
Now every developer in the team must manually run this ktlint command 
`./ktlint --apply-to-idea-project --android`

This is fine, but I wonder can it be achieved without separatley downloading (and keeping) ktlint binaries? I would be nice to provide ability to configure IDE out of the box.

hmm Maybe plugin could somehow link `./ktlint` command to project root or add new gradle task that would allow passing parameters to ktlint script `./gradlew ktlint` 🤔

(we could, of course, add new grade task but I feel this would not be flexible enough if new ktlint parameters will be added in future)

Tapchicoma:
This task will generate formatting rules for IDEA. It will overwrite
existing ones if they already are present.

Closes #91.


JLLeitschuh:
I really like this idea.

@nedtwigg what are your thoughts on adding something similar to spotless?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -28,8 +28,10 @@ import kotlin.reflect.KClass
 
 const val VERIFICATION_GROUP = "Verification"
 const val FORMATTING_GROUP = "Formatting"
+const val HELP_GROUP = "Help"
```
JLLeitschuh: I think a few of these already exist as constants in the Gradle public API. Any thoughts on simply using them instead?

nedtwigg:
I'm ignorant about IntelliJ, but I'm always happy to merge things which are useful to their authors :)

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: Wasn't aware of them, can you point where I can find them?

Tapchicoma: So far only found in `HelpTasksPlugin.HELP_GROUP`

Tapchicoma: I've add using `HelpTaskPlugin.HELP_GROUP` constant.

JLLeitschuh: `LifecycleBasePlugin.VERIFICATION_GROUP`

Tapchicoma: added

Tapchicoma:
@JLLeitschuh I think it is ready to be merged

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -265,6 +284,26 @@ open class KtlintPlugin : Plugin<Project> {
...
+    private fun addApplyToIdeaTask(target: Project, extension: KtlintExtension) {
+        target.afterEvaluate {
+            val rootProject = target.rootProject
+            if (rootProject.tasks.findByName(APPLY_TO_IDEA_TASK_NAME) == null) {
+                val ktLintConfig = createConfiguration(rootProject, extension)
+                // -y here to auto-overwrite existing IDEA code style
+                val runArgs = mutableSetOf("--apply-to-idea-project", "-y")
+                addAndroidArg(extension, runArgs)
+
+                target.rootProject.taskHelper<JavaExec>("ktlintApplyToIDEA") {
```
JLLeitschuh: This is kinda inconsistent with the gradle `idea` task naming scheme.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -265,6 +284,26 @@ open class KtlintPlugin : Plugin<Project> {
...
+                    group = HELP_GROUP
+                    description = "Generates IDEA built-in formatter rules. It will overwrite existing ones."
+                    main = "com.github.shyiko.ktlint.Main"
+                    classpath = ktLintConfig
+                    args(runArgs)
+                }
```
JLLeitschuh: Any reason why you chose to make this a `JavaExec` task instead of standalone task?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -166,20 +172,33 @@ open class KtlintPlugin : Plugin<Project> {
...
+        runArgs: MutableSet<String>
+    ) {
         if (extension.android && SemVer.parse(extension.version) >= SemVer(0, 12, 0)) {
             // Android option is available from ktlint 0.12.0
             runArgs.add("--android")
         }
-        if (extension.ruleSets.size > 0) {
-            extension.ruleSets.forEach { runArgs.add("--ruleset=$it") }
-        }
     }
```
JLLeitschuh: General question, why is a mutable collection passed through all of this code instead of having each method append all of these list together as each method returns?

It seems weird to have a mutable list floating through these various methods.

Not related to your code changes directly, I just realized that this was kind of a weird practice going on .

JLLeitschuh:
Sorry, I'm terrible for having left you with this PR without quality feedback for so long. I've been scatter brained about this. I'll try to be better about this in the future.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: We still have format task as a `JavaExec`, but can make it standalone as well.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: addressed

JLLeitschuh:
Awesome! Can you call out this task exists in the README?

README.md
```diff
@@ -83,6 +83,10 @@ Similarly, a `ktlintFormat` task has been added that formats all of the source s
 If project has subprojects - plugin also adds two meta tasks `ktlintCheck` and `ktlintFormat` to the root project that
 triggers related tasks in subprojects.
 
+Another task added: `ktlintApplyToIdea`. It is always added only to the
+root project. Task generates IntelliJ IDEA (or Android Studio) Kotlin
+style files in `.idea/` folder. **Note** that this task will *overwrite*
+existing ones style files.
```
JLLeitschuh: > **Note** that this task will *overwrite* existing ones style files.

Gramatically, that's weird.

Tapchicoma: English is not my native language. What is the right wording? *will replace*?

JLLeitschuh: Perhaps:
> Note that this task will overwrite any existing style file.

Unless there is a singular file that it could overwritten, then it might be:

> Note that this task will overwrite the existing style file.

Tapchicoma: fixed the wording

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -185,4 +185,15 @@ class KtlintPluginTest : AbstractPluginTest() {
...
         }
     }
+
+    @Test
+    fun `Should generate code style files`() {
+        projectRoot.withCleanSources()
+        val ideaRootDir = projectRoot.resolve(".idea").apply { mkdir() }
+
+        build("ktlintApplyToIDEA").apply {
+            assertThat(task(":ktlintApplyToIdea")!!.outcome, equalTo(TaskOutcome.SUCCESS))
```
igorwojda: It would be better to do this, so the test will fail properly (throwing proper assertion framework exception that shows exactly what is wrong, instead of `KotlinNullPointerException`)
`task(":ktlintApplyToIdea")?.outcome`

PR #99, opened 2018-06-05
WIP: Use android docker image on Travis CI.

Tapchicoma:
This should improve build time, as this image will already contain
all required android build tools.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,18 @@
+sudo: required
+
 language: android
-jdk:
-  - oraclejdk8
-install:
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Doesn't `sudo: required` make travis use their older architecture which is inherently slower?

Tapchicoma:
I am just checking if it worth to use. My intention was if docker image is cached - then build should be fast enough. If not - will be approx. the same time.

Tapchicoma:
Closing until better docker support (with images caching) on travis.

PR #104, opened 2018-06-17
Configure publishing via Travis CI

JLLeitschuh:
Once a release commit has been pushed, you can simply checkout the `master` branch and run.
```
git tag v[whatever version]
```
From there run `git push --tags` and travis CI will publish the release for you.

Note: Gradle still reads the version from the `version` variable, not from git.
It's expected that these two versions are the same when a release is performed.

Also, I haven't been able to test this yet as it requires performing a release.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -51,6 +51,29 @@ publishing {
     }
 }
 
+/**
+ * Configures the publishing environment for publishing with Travis CI.
+ * All you need to do is push a tagged commit to github and Travis CI will automatically publish a
+ * release of the plugin using the current [Project.getVersion].
+ */
+fun setupPublishingEnvironment() {
+    val keyEnvironmentVariable = "GRADLE_PUBLISH_KEY"
+    val secretEnvironmentVariable = "GRADLE_PUBLISH_SECRET"
+
+    val keyProperty = "gradle.publish.key"
+    val secretProperty = "gradle.publish.secret"
+
+    if (System.getProperty(keyProperty) == null || System.getProperty(secretProperty) == null) {
+        logger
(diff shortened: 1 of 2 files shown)


plugin/build.gradle.kts
```diff
@@ -51,6 +51,29 @@ publishing {
...
+    val secretEnvironmentVariable = "GRADLE_PUBLISH_SECRET"
+
+    val keyProperty = "gradle.publish.key"
+    val secretProperty = "gradle.publish.secret"
+
+    if (System.getProperty(keyProperty) == null || System.getProperty(secretProperty) == null) {
+        logger
+            .info("`$keyProperty` or `$secretProperty` were not set. Attempting to configure from environment variables")
+
+        System.setProperty(keyProperty, System.getProperty(keyEnvironmentVariable))
```
Tapchicoma: This is the problem on CI:
```kotlin
Caused by: java.lang.NullPointerException
        at Build_gradle.setupPublishingEnvironment(build.gradle.kts:70)
        at Build_gradle.<init>(build.gradle.kts:75)
```

Tapchicoma: This is caused when `GRADLE_PUBLISH_KEY` or `GRADLE_PUBLISH_SECRET` is unset.

JLLeitschuh: Thanks for figuring out where this was coming from. I haven't had time to fix this.
Feel free to resolve this if you have the time. Otherwise, I'll try to get to this later this week.

Tapchicoma:
LGTM!

JLLeitschuh:
Tell me if it doesn't work for the next release and I'll fix it.

PR #108, opened 2018-06-26
New release cycle

Tapchicoma:
Also updated several dependencies versions.


gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.configureondemand=false
```
JLLeitschuh: Issues with the Kotlin DSL?

Tapchicoma: No, issue with Android gradle plugin and new Gradle wrapper versions. Last version of Gradle wrapper that worked with such configuration was `4.5.1`.

JLLeitschuh: Is there a known bug that we should link to above this? Just so we have some bug we can check to see when we can re-enable this feature?

Tapchicoma: https://issuetracker.google.com/issues/76429932 - according to the issue it will be fixed in next android plugin release: `3.2` (currently in beta)

PR #112, opened 2018-08-06
Fix Auto-Release

JLLeitschuh:
Unfortunately, this is still not 100% correct.
We are still creating two different travis jobs, one for each secure variable.
Instead I want to create one travis job with the two secure variables.

This stack overflow issue sums up my problem pretty well.
https://stackoverflow.com/questions/45993334/multiple-values-of-a-secure-encrypted-environment-variable-in-travis-yml


Tapchicoma:
I am not very familiar with Travis, but I assume we are currently has only one job.

PR #115, opened 2018-08-21
Fix .editorconfig file change doesn't reset check task state.

### LINKED ISSUE #106: Editorconfig file should be treated as input
Currently check task caching doesn't take into consideration any changes in `.editorconfig` file. 

For example, if I add `max_line_length=40` to `.editorconfig`, task result will still be taken from Gradle cache.

Tapchicoma:
This PR fixes issue that `ktlintCheck` task ignores `.editorconfig` file changes.

For now task assumes that `.editorconfig` file will be either in root project folder or in current project folder. If somebody will need more advanced logic - new issue should be created and this problem will be addressed in a separate PR.

Fixes #106.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
@@ -44,10 +44,10 @@ open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : Defau
         }
     }
 
-    @get:SkipWhenEmpty
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:InputFiles
     internal val editorConfigFiles: FileCollection = getEditorConfigFiles(project)
+
     @get:Input
     val verbose: Property<Boolean> = objectFactory.booleanProperty()
     @get:Input
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
@@ -4,6 +4,7 @@ import net.swiftzer.semver.SemVer
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileTree
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
@@ -25,6 +26,8 @@ import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
 import java.io.File
 import javax.inject.Inject
 
+private const val EDITORCONGIG_FILE_NAME = ".editorconfig"
+
 @CacheableTask
 open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : DefaultTask() {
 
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -43,6 +46,11 @@ open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : Defau
         }
     }
 
+    @get:SkipWhenEmpty
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    @get:InputFiles
+    val editorConfigFiles: FileCollection = project.rootProject.files(EDITORCONGIG_FILE_NAME)
+        .plus(project.files(EDITORCONGIG_FILE_NAME))
```
JLLeitschuh: Do you want to use a `RegularFileProperty` instead? Then you can annotate it as `@get:InputFile` instead of `@get:InputFiles`. Unless you think the user can specify multiple files.

Also, do you want to consider making this field `internal` or do you want to allow users to modify it?

JLLeitschuh:
Thanks for working on fixing this!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
JLLeitschuh: Totally makes sense.

Another problem with the way you have this is that the root project is not always necessarily in a parent directory.
It is completely valid to have a structure like this:
```
someDir
 - rootProject
   - rootProject.gradle
   - .editorconfig
   - settings.gradle
 - subprojects
   - subprojects.gradle
   - .editorconfig
     - subprojectA
       - subprjoectA.gradle
       - .editorconfig
```
I understand this is an atypical project structure. But it is completely valid.

See the "Where are these files stored?" section from this document to see how the filesystem walking should really happen:
https://editorconfig.org/

Tapchicoma: Ok, I will try to adapt proposed strategy for detecting `.editorconfig` files that is seems to be similar how it is in the Ktlint :+1:

Tapchicoma:
@JLLeitschuh reworked how `.editorconfig` files are detected - it uses similar algorithm as KtLint.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -43,6 +44,10 @@ open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : Defau
         }
     }
 
+    @get:SkipWhenEmpty
```
JLLeitschuh: Uhhhh. I don't think we want to skip the task when there isn't a `.editorconfig` file.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
```diff
@@ -43,6 +44,10 @@ open class KtlintCheck @Inject constructor(objectFactory: ObjectFactory) : Defau
...
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    @get:InputFiles
+    internal val editorConfigFiles: FileCollection = getEditorConfigFiles(project)
     @get:Input
```
JLLeitschuh: Can you add a newline here above this line.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -22,6 +24,56 @@ internal inline fun <reified T : Task> Project.taskHelper(name: String, noinline
...
+    return if (parentDir != null &&
+        parentDir != project.rootDir.toPath() &&
+        !editorConfigFC.isRootEditorConfig()) {
+        searchEditorConfigFiles(project, parentDir, outputCollection)
+    } else {
+        outputCollection
+    }
+}
+
+private val editorConfigRootRegex = "^root\\s?=\\s?true\\n".toRegex()
```
JLLeitschuh: Thank you for caching this so it only needs to be compiled once! 👍

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheck.kt
Tapchicoma: :scream: Thank you for pointing this out!

Tapchicoma:
@JLLeitschuh fixed comments

PR #120, opened 2018-09-02
Add new expermental kotlin native plugin support

### LINKED ISSUE #119: Plugin doesn't create tasks in Kotlin/Native modules
In a multiplatform project that adds the plugin via `subprojects` (here's the [PR](https://github.com/square/okio/pull/502) that adds ktlint), tasks are not being created for the Kotlin/Native module. Running

```
./gradlew :okio:js:ktlintCheck
```

succeeds, while

```
./gradlew :okio:native:ktlintCheck
```

produces

```
* What went wrong:
Task 'ktlintCheck' not found in project ':okio:native'.
```

Looking at the code, I see that the plugin is supposed to work with Kotlin/Native, can try to debug later to see what's going wrong.

Tapchicoma:
Adds support for new kotlin native experimental Gradle plugin: https://github.com/JetBrains/kotlin-native/blob/master/GRADLE_PLUGIN.md#experimental-plugin

Closes #119.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -135,19 +140,74 @@ open class KtlintPlugin : Plugin<Project> {
...
 
-                        val ktlintSourceSetFormatTask = createFormatTask(project, it.name, ktLintConfig,
-                                kotlinSourceSet, runArgs)
+                        val ktlintSourceSetFormatTask = createFormatTask(
+                            project,
+                            component.name,
+                            ktLintConfig,
+                            kotlinSourceSet,
+                            runArgs
+                        )
```
JLLeitschuh: Anything you can do to de-duplicate this code a bit more?

JLLeitschuh: It seems like this is the first time you're doing the duplication. Feel free to tackle it now instead of a seperate PR, even if the diff is more ugly, it will help with future maintainability.

samples/kotlin-native/src/main/kotlin/org/jlleitschuh/gradle/ktlint/sample/native/main.kt
```diff
@@ -1,3 +1,5 @@
+package org.jlleitschuh.gradle.ktlint.sample.native
```
JLLeitschuh: 👍

samples/kotlin-native/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ kotlinNativeSourceSets["main"].component {
     if (this is KotlinNativeMainComponent) {
         outputKinds.set(listOf(OutputKind.EXECUTABLE))
         baseName.set("foo")
+        extraOpts(listOf("-entry", "org.jlleitschuh.gradle.ktlint.sample.native.main"))
```
JLLeitschuh: Why this change?

Tapchicoma: New experimental plugin has not yet supported defining entry point, like "konan" plugin. 

There is open PR: https://github.com/JetBrains/kotlin-native/pull/1996 that adds such support.

JLLeitschuh: 👍

Issue #111, opened 2018-08-01
KtlintFormat should be it's own task


JLLeitschuh:
All of the other tasks are currently their own tasks.
`ktlintFormat` is the only task left that is still a `JavaExec` task.

PR #128, opened 2018-09-08
Remove usages of afterEvaluate {}

### LINKED ISSUE #122: Consider removing afterEvaluate
Please consider removing afterEvaluate from the plugin configuration scripts. 

First of all, overusing afterEvaluate is bad for the user. When a plugin uses it, it could potentially result into bigger problems. 

Instead more reactive/live collections should be used to detect the usages in the project and react accordingly. Example is:

```
project.plugins.withId('kotlin') {
    configureKotlinProject()
}
```

Here we are able to configure the project just after the kotlin plugin is applied. 

Second problem I faced is regarding another plugin I contribute which configures other static analysis plugins: https://github.com/novoda/gradle-static-analysis-plugin/issues/129

Because of the `afterEvaluate` usage, I also needed to use `afterEvaluate` in that project and that results into nondeterministic behavior. Depending on the client project I integrate, I started to get task not found problems. I think I can resolve this particular issue in `static-analysis-plugin`.

But still I believe that everything will be nicer without `afterEvaluate`.

Tapchicoma:
This should improve plugin configuration performance and fix deprecation issue in Gradle `4.10`.

Closes #122 and #111

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -120,7 +120,7 @@ open class KtlintCheckTask @Inject constructor(
 
     private fun checkMinimalSupportedKtlintVersion() {
         if (SemVer.parse(ktlintVersion.get()) < SemVer(0, 10, 0)) {
-            throw GradleException("Ktlint versions less than 0.10.0 are not supported." +
+            throw GradleException("Ktlint versions less than 0.10.0 are not supported. " +
                 "Detected Ktlint version: ${ktlintVersion.get()}.")
         }
     }
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -120,7 +120,8 @@ open class KtlintCheckTask @Inject constructor(
 
     private fun checkMinimalSupportedKtlintVersion() {
         if (SemVer.parse(ktlintVersion.get()) < SemVer(0, 10, 0)) {
-            throw GradleException("Ktlint versions less than 0.10.0 are not supported. Detected Ktlint version: $this.")
+            throw GradleException("Ktlint versions less than 0.10.0 are not supported." +
+                "Detected Ktlint version: ${ktlintVersion.get()}.")
         }
     }
 
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -21,6 +21,7 @@ import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
 import org.gradle.api.tasks.SkipWhenEmpty
 import org.gradle.api.tasks.TaskAction
+import org.gradle.process.JavaExecSpec
 import org.jlleitschuh.gradle.ktlint.reporter.KtlintReport
 import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
 import javax.inject.Inject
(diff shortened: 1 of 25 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -24,6 +25,24 @@ internal inline fun <reified T : Task> Project.taskHelper(name: String, noinline
...
 
+internal inline fun <reified T : Task> Project.maybeCreateTask(
+    name: String,
+    noinline configuration: T.() -> Unit
+): T {
+    return tasks.maybeCreate(name, T::class.java).apply { configure(closureOf(configuration)) }
+}
+
+// Port from Kotlin DSL
+fun <T> Any.closureOf(action: T.() -> Unit): Closure<Any?> = KotlinClosure1(action, this, this)
```
Tapchicoma: @JLLeitschuh Not sure that I did it right here, but IDE failed to find `closureOf {}`

JLLeitschuh: I'm not totally certain why you are choosing to use `maybeCreateTask`..
I don't think that anyone else will be creating these tasks, and if they are, then this plugin should fail loudly.

JLLeitschuh:
You need to check and make sure we don't touch variables on the `Extension` anywhere that isn't lazily evaluated.
Doing anything else will cause us to be working with data that the user may not have configured as their "final" configuration.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
Tapchicoma: You are right, `maybeCreateTask` is not needed in this case.

Tapchicoma:
@JLLeitschuh what is the right way to access `KtlintExtention` with final configuration? Currently plugin access values of extention everywhere whenever it creates the tasks.

JLLeitschuh:
@Tapchicoma That sounds like a reasonable solution, or you could fail the task in the `@TaskAction` annotated method.

I think that makes sense.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -47,5 +54,7 @@ open class KtlintExtension {
      *
      * Default is plain.
      */
-    var reporters: Array<ReporterType> = arrayOf(ReporterType.PLAIN)
+    val reporters: SetProperty<ReporterType> = objectFactory.setProperty {
```
Tapchicoma: @Jlleitschuh Changing this to `SetProperty<ReporterType> bring problem that in common gradle script it is not allowed anymore to use `reporters = ["CHECKSTYLE", "PLAIN"]` and I don't know for now how address it better.

See example how it looks now here: https://github.com/JLLeitschuh/ktlint-gradle/blob/094995594f19bf25a2f4d94797afb5f72a09b317/samples/kotlin-gradle/build.gradle#L16

Tapchicoma:
@JLLeitschuh ready to be review, when you will have time

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.gradle.api.tasks.TaskAction
+import org.gradle.process.JavaExecSpec
+import org.jlleitschuh.gradle.ktlint.reporter.KtlintReport
+import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
+import javax.inject.Inject
+
+@CacheableTask
+open class KtlintCheckTask @Inject constructor(
+    private val objectFactory: ObjectFactory
+) : DefaultTask() {
```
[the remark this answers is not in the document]
Tapchicoma: After playing with styling, I think - I will stick to the current one. Having multiple annotations for the class definintion a little disturbes me:
```kotlin
@CacheableTask
open class KtlintCheckTask 
@Inject 
constructor(
    private val objectFactory: ObjectFactory
) : DefaultTask() {
  ...
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -0,0 +1,151 @@
...
+    private fun generateJavaExecSpec(
+        additionalConfig: (JavaExecSpec) -> Unit
+    ): (JavaExecSpec) -> Unit = { javaExecSpec ->
+        javaExecSpec.classpath = classpath
+        javaExecSpec.main = "com.github.shyiko.ktlint.Main"
+        javaExecSpec.args(sourceDirectories.flatMap { dir ->
+            KOTLIN_EXTENSIONS.map { extension ->
+                "${dir.path}/**/*.$extension"
+            }
+        })
```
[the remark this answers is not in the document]
Tapchicoma: I think this case exists for the mixed kotlin/java code in the same sources.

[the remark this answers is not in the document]
Tapchicoma: Also ktlint in examples uses such approach:
```shell
# check only certain locations (prepend ! to negate the pattern) 
$ ktlint "src/**/*.kt" "!src/**/*Test.kt"
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -0,0 +1,151 @@
...
+    val android: Property<Boolean> = objectFactory.property()
+    @get:Input
+    val ignoreFailures: Property<Boolean> = objectFactory.property()
+    @get:Internal
+    val reporters: SetProperty<ReporterType> = objectFactory.setProperty()
+    @get:Console
+    val outputToConsole: Property<Boolean> = objectFactory.property()
+
+    @get:Internal
+    val enabledReports
```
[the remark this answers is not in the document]
Tapchicoma: Change visibility to `internal` in 59fb622.

Left public visibility only for `reportOutputFiles` that may be used by plugin consumers to collect all reports in one place.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: Possible solution is to use `SetProperty<String>` in extention and convert it to `SetProperty<ReporterType>` when passing to the task.

On the other hand, now, it is more obvious what reporter types are available, and has the same approach as kotlin dsl extention configuration.

@JLLeitschuh what do you think about it?

PR #129, opened 2018-09-20
Remove custom wrapper task.

Tapchicoma:
This approach is deprecated in Gradle.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,7 +27,7 @@ apply {
     plugin("org.jlleitschuh.gradle.ktlint")
 }
 
-task<Wrapper>("wrapper") {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -27,7 +27,7 @@ apply {
     plugin("org.jlleitschuh.gradle.ktlint")
 }
 
-task<Wrapper>("wrapper") {
+tasks.withType(Wrapper::class.java) {
```
JLLeitschuh: @Tapchicoma In the newest version of gradle, the correct way to do this in a way that defers configuration is to do:
```kotlin
tasks.withType(Wrapper::class.java).configureEach { // ...
```

PR #135, opened 2018-09-29
Fix plugin fails to configure android projects with flavors.

### LINKED ISSUE #131: Compatibility issue with Android tools 3.3.0-alpha11
I just tried the `6.0.0` version of ktlint Gradle plugin with the latest version of Android Gradle plugin. 

The latest version of ktlint removed the `afterEvaluate`
The latest version of AGP introduced lazy task creation. 

When they come together, I got the following error: 

> Cannot add task ‘assembleDebug’ as a task with that name already exists.

<details>
Caused by: org.gradle.api.internal.tasks.DefaultTaskContainer$DuplicateTaskException: Cannot add task 'assembleDebug' as a task with that name already exists.
    at org.gradle.api.internal.tasks.DefaultTaskContainer.duplicateTask(DefaultTaskContainer.java:250)
    at org.gradle.api.internal.tasks.DefaultTaskContainer.registerTask(DefaultTaskContainer.java:365)
    at org.gradle.api.internal.tasks.DefaultTaskContainer.register(DefaultTaskContainer.java:345)
    at com.android.build.gradle.internal.tasks.factory.TaskFactoryImpl.register(TaskFactoryImpl.kt:84)
    at com.android.build.gradle.internal.TaskManager.createAnchorAssembleTasks(TaskManager.java:3038)
    at com.android.build.gradle.BasePlugin.createAndroidTasks(BasePlugin.java:814)
    at com.android.builder.profile.ThreadRecorder.record(ThreadRecorder.java:81)
    at com.android.build.gradle.BasePlugin.lambda$createTasks$4(BasePlugin.java:667)
    at com.android.build.gradle.internal.crash.CrashReporting$afterEvaluate$1.execute(crash_reporting.kt:37)
    at com.android.build.gradle.internal.crash.CrashReporting$afterEvaluate$1.execute(crash_reporting.kt)
</details>

AGP tries to `register` the task but the `assembleDebug` task is already added by `ktlint` Gradle plugin. Not sure if `ktlint` really creates a task named `assembleDebug` but it would worth double-checking. 

Reverting back to `5.0.0` fixes the problem. :+1:

Tapchicoma:
Rethink approach how plugin gets android project sources and now it
uses android plugins base extension information, rather then VariantManager.

This also removes check of the same code more then once.

Closes #131

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -111,9 +111,11 @@ open class KtlintPlugin : Plugin<Project> {
                 addKtlintFormatTaskToProjectMetaFormatTask(target, ktlintSourceSetFormatTask)
             }
 
-            // Variant manager returns all sources for variant,
-            // so most probably main source set maybe checked several times.
-            // This approach creates one check tasks per one source set.
+            /*
+            Variant manager returns all sources for variant,
+            so most probably main source set maybe checked several times.
+            This approach creates one check tasks per one source set.
+            */
             val pluginConfigure: (Plugin<Any>) -> Unit = { _ ->
                 target.extensions.configure(BaseExtension::class.java) { ext ->
                     ext.sourceSets { sourceSet ->
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -116,7 +116,7 @@ open class KtlintPlugin : Plugin<Project> {
             so most probably main source set maybe checked several times.
             This approach creates one check tasks per one source set.
             */
-            val pluginConfigure: (Plugin<Any>) -> Unit = { _ ->
+            fun getPluginConfigureAction(): (Plugin<Any>) -> Unit = { _ ->
                 target.extensions.configure(BaseExtension::class.java) { ext ->
                     ext.sourceSets { sourceSet ->
                         sourceSet.all {
@@ -126,11 +126,11 @@ open class KtlintPlugin : Plugin<Project> {
                 }
             }
 
-            target.plugins.withId("com.android.application", pluginConfigure)
-            target.plugins.withId("com.android.library", pluginConfigure)
-            target.plugins.withId("com.android.instantapp", pluginConfigure)
-            target.plugins.withId("com.android.feature", pluginConfigure)
-            target.plugins.withId("com.android.test", pluginConfigure)
+            target.plugins.withId("com.android.application", getPluginConfigureAction())
+            target.plugins.withId("com.android.library", getPluginConfigureAction())
+            target.plugins.withId("com.android.instantapp", getPluginConfigureAction())
+            target.plugins.withId("com.android.feature", getPluginConfigureAction())
+            target.plugins.withId("com.android.test", getPluginConfigureAction())
         }
     }
 
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -112,10 +112,10 @@ open class KtlintPlugin : Plugin<Project> {
             }
 
             /*
-            Variant manager returns all sources for variant,
-            so most probably main source set maybe checked several times.
-            This approach creates one check tasks per one source set.
-            */
+             * Variant manager returns all sources for variant,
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -91,56 +85,50 @@ open class KtlintPlugin : Plugin<Project> {
...
-            target.plugins.withId("com.android.feature") { plugin ->
-                (plugin as FeaturePlugin).variantManager.run(createTasks)
+                addKtlintFormatTaskToProjectMetaFormatTask(target, ktlintSourceSetFormatTask)
             }
-            target.plugins.withId("com.android.test") { plugin ->
-                (plugin as TestPlugin).variantManager.run(createTasks)
+
+            // Variant manager returns all sources for variant,
+            // so most probably main source set maybe checked several times.
+            // This approach creates one check tasks per one source set.
```
JLLeitschuh: Multiline comment instead?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -91,56 +85,50 @@ open class KtlintPlugin : Plugin<Project> {
...
+            val pluginConfigure: (Plugin<Any>) -> Unit = { _ ->
+                target.extensions.configure(BaseExtension::class.java) { ext ->
+                    ext.sourceSets { sourceSet ->
+                        sourceSet.all {
+                            createTasks(it.name, target.files(it.java.srcDirs))
+                        }
+                    }
+                }
             }
```
JLLeitschuh: Why not make this a function? This is fine if it's just straight up simpler.

samples/android-app/build.gradle.kts
```diff
@@ -26,6 +26,12 @@ configure<AppExtension> {
         }
     }
 
+    flavorDimensions("beer")
+    productFlavors {
+        register("weissbier")
+        register("kellerbier")
+    }
```
JLLeitschuh: Can you explain this?

Tapchicoma: It is android build variants. 

For example, in this case, app can include special sources/resources that only belongs to `"weissbier"` variant on top of main sources/resources.

Such approach allows to have multiple variants of the same app, like special variant for play store and variant of the app for F-Droid store.

JLLeitschuh: Cool, good to know!

Tapchicoma:
@JLLeitschuh addressed comments

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -111,10 +111,12 @@ open class KtlintPlugin : Plugin<Project> {
...
 
-            // Variant manager returns all sources for variant,
-            // so most probably main source set maybe checked several times.
-            // This approach creates one check tasks per one source set.
-            val pluginConfigure: (Plugin<Any>) -> Unit = { _ ->
+            /*
+            Variant manager returns all sources for variant,
+            so most probably main source set maybe checked several times.
+            This approach creates one check tasks per one source set.
+            */
```
JLLeitschuh: This is more what I meant:
```kotlin
            /*
             * Variant manager returns all sources for variant,
             * so most probably main source set maybe checked several times.
             * This approach creates one check tasks per one source set.
             */
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -124,11 +126,11 @@ open class KtlintPlugin : Plugin<Project> {
...
-            target.plugins.withId("com.android.application", pluginConfigure)
-            target.plugins.withId("com.android.library", pluginConfigure)
-            target.plugins.withId("com.android.instantapp", pluginConfigure)
-            target.plugins.withId("com.android.feature", pluginConfigure)
-            target.plugins.withId("com.android.test", pluginConfigure)
+            target.plugins.withId("com.android.application", getPluginConfigureAction())
+            target.plugins.withId("com.android.library", getPluginConfigureAction())
+            target.plugins.withId("com.android.instantapp", getPluginConfigureAction())
+            target.plugins.withId("com.android.feature", getPluginConfigureAction())
+            target.plugins.withId("com.android.test", getPluginConfigureAction())
```
JLLeitschuh: Uhh. Sure. this works 😆 Thanks!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: Changed to this :slightly_smiling_face:

PR #136, opened 2018-09-29
Hide specific source sets tasks.

Tapchicoma:
Meta tasks (`ktlintCheck` and `ktlintFormat`) in the project are still visible.

My intention is to reduce unnessesary information, as usually ktlint used by running just `ktlintCheck` (or `ktlintFormat`) tasks.

This hidden tasks can still be visible by running `./gradlew tasks --all`.


JLLeitschuh:
Can you add a test so we ensure we keep this behaviour in the future?
It will ensure this doesn't unintentionally get reverted in a future refactor.

Tapchicoma:
@JLLeitschuh added tests

PR #137, opened 2018-09-29
Add support for console colored output.

### LINKED ISSUE #50: Add support for colored output
Ktlint version `0.13.0` introduced colored ouput support:

> --color CLI option for colored output (where supported, e.g. --print-ast, default (plain) reporter, etc)

Tapchicoma:
Closes #50

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
@@ -34,6 +34,10 @@ open class KtlintExtension(
      */
     val outputToConsole: Property<Boolean> = objectFactory.property { set(true) }
     /**
+     * Enabled colored output to console.
+     */
+    val coloredOutput: Property<Boolean> = objectFactory.property { set(false) }
+    /**
      * Whether or not to allow the build to continue if there are warnings;
      * defaults to {@code false}, as for any other static code analysis tool.
      * <p>
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
@@ -36,7 +36,7 @@ open class KtlintExtension(
     /**
      * Enabled colored output to console.
      */
-    val coloredOutput: Property<Boolean> = objectFactory.property { set(false) }
+    val coloredOutput: Property<Boolean> = objectFactory.property { set(true) }
     /**
      * Whether or not to allow the build to continue if there are warnings;
      * defaults to {@code false}, as for any other static code analysis tool.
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -33,6 +33,10 @@ open class KtlintExtension(
      * Enable console output mode.
      */
     val outputToConsole: Property<Boolean> = objectFactory.property { set(true) }
+    /**
+     * Enabled colored output to console.
+     */
+    val coloredOutput: Property<Boolean> = objectFactory.property { set(false) }
```
JLLeitschuh: Why default to false? I'm good either way on this one, but why make this the default?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -271,6 +273,9 @@ open class KtlintPlugin : Plugin<Project> {
         android.set(extension.android)
         ignoreFailures.set(extension.ignoreFailures)
         outputToConsole.set(extension.outputToConsole)
+        if (!target.isConsolePlain()) {
+            coloredOutput.set(extension.coloredOutput)
```
JLLeitschuh: You should probably do this in a `extension.coloredOutput.map { }` where you put in the debug log that you are disabling the user choice because of the `isConsolePlain()` flag.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: Yes, I think default value as `true` will be better, as on plain console type it will not be active.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -273,9 +273,14 @@ open class KtlintPlugin : Plugin<Project> {
         android.set(extension.android)
         ignoreFailures.set(extension.ignoreFailures)
         outputToConsole.set(extension.outputToConsole)
-        if (!target.isConsolePlain()) {
-            coloredOutput.set(extension.coloredOutput)
-        }
+        coloredOutput.set(extension.coloredOutput.map {
+            if (target.isConsolePlain()) {
+                target.logger.info("Console type is plain: disabling colored output")
```
JLLeitschuh: `info` isn't logged by default correct? If so 👍

Tapchicoma: Yes, by default warn and error is logged.

PR #147, opened 2018-10-10
Allow to exclude sources from check

### LINKED ISSUE #97: Provide an extension for excluding directories
I'd like to be able to ignore the build directory for lint tasks

```
afterConfigure {
  ktlintMainCheck {
    exclude "build"
  }
}
```

Tapchicoma:
Closes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,9 +154,10 @@ ktlint {
         "/path/to/custom/rulseset.jar",
         "com.github.username:rulseset:master-SNAPSHOT"
     ]
-    exclude = [
-        "**/generated/**"
-    ]
+    filter {
+        exclude("**/generated/**")
+        include("**/kotlin/**")
+    }
 }
 ```
 
@@ -176,7 +177,10 @@ ktlint {
         "/path/to/custom/rulseset.jar",
         "com.github.username:rulseset:master-SNAPSHOT"
     ))
-    exclude.set(setOf("**/generated/**")
+    filter {
+        exclude("**/generated/**")
+        include("**/kotlin/**")
+    }
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -154,6 +154,9 @@ ktlint {
         "/path/to/custom/rulseset.jar",
         "com.github.username:rulseset:master-SNAPSHOT"
     ]
+    exclude = [
+        "**/generated/**"
+    ]
 }
 ```
 
@@ -173,6 +176,7 @@ ktlint {
         "/path/to/custom/rulseset.jar",
         "com.github.username:rulseset:master-SNAPSHOT"
     ))
+    exclude.set(setOf("**/generated/**")
 }
 ```
 
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
@@ -4,6 +4,7 @@ import org.gradle.api.Action
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternFilterable
+import org.gradle.util.GradleVersion
 
 /**
  * The base Ktlint plugin that all other plugins are built on.
@@ -13,7 +14,12 @@ open class KtlintBasePlugin : Plugin<Project> {
 
     override fun apply(target: Project) {
         val filterTargetApplier = fun(action: Action<PatternFilterable>) {
-            target.tasks.withType(KtlintCheckTask::class.java).configureEach(action)
+            if (GradleVersion.current() >= GradleVersion.version("4.9")) {
+                // API only added in Gradle 4.9
+                target.tasks.withType(KtlintCheckTask::class.java).configureEach(action)
+            } else {
+                target.tasks.withType(KtlintCheckTask::class.java, action)
+            }
         }
         extension = target.extensions.create(
             "ktlint",
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
@@ -6,7 +6,7 @@ import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternFilterable
 import org.gradle.util.GradleVersion
 
-typealias FilterApplier = (Action<PatternFilterable>) -> Unit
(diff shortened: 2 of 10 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -57,6 +57,8 @@ open class KtlintCheckTask @Inject constructor(
     internal val outputToConsole: Property<Boolean> = objectFactory.property()
     @get:Console
     internal val coloredOutput: Property<Boolean> = objectFactory.property()
+    @get:Internal
```
Tapchicoma: Not so sure if `Internal` here is good choice, as it exclude sources in task action.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -61,4 +61,12 @@ open class KtlintExtension(
...
         set(setOf(ReporterType.PLAIN))
     }
+
+    /**
+     * Exclude sources under given source patterns.
+     *
+     * See [PatternFilterable](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html)
+     * for details how source patterns should look.
+     */
+    val exclude: SetProperty<String> = objectFactory.setProperty()
```
JLLeitschuh: I'd rather you just expose a method `fun matching( filterConfigAction: Action<PatternFilterable>)` instead.
That way any consumer can specify a custom matching pattern for inclusion/exclusion at their leisure.

Tapchicoma: Do I understand it right - you want to have in extension something like this?

```kotlin
    internal var filter: Action<PatternFilterable> = Action {} 
    fun matching(filter: Action<PatternFilterable>) {
        this.filter = filter
    }
```

JLLeitschuh: I think that makes sense. Do you?

Tapchicoma: I am a little worry about allowing to use `include()` and possible misunderstanding from users, as they may think that they should include manually all kotlin sources.

On the other hand I see a benefit as well of such approach.

JLLeitschuh: They can do the same thing if they use `tasks.withType<KtlintCheckTask>().configureEach {`

Tapchicoma: valid point :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -72,6 +73,9 @@ open class KtlintCheckTask @Inject constructor(
             }
             .filter { it.enabled.get() }
 
+    @get:Nested
+    internal val filterAction: Property<Action<PatternFilterable>> = objectFactory.property()
```
JLLeitschuh: 🤔 I.... Hrmmm...
I don't think this should be a field, I think it should be applied by the plugin logic before the `@TaskAction` is evaluated.

If the `filterAction`changes, the `source` will change thus invalidating the task.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -62,11 +64,16 @@ open class KtlintExtension(
...
 
+    internal val filterAction: Property<Action<PatternFilterable>> = objectFactory
+        .property { set(Action {}) }
+
     /**
-     * Exclude sources under given source patterns.
+     * Filters sources using given source patterns.
      *
      * See [PatternFilterable](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html)
      * for details how source patterns should look.
```
JLLeitschuh: This doc is no longer relevant because it's not a "pattern" anymore.

README.md
```diff
@@ -176,7 +177,10 @@ ktlint {
         "/path/to/custom/rulseset.jar",
         "com.github.username:rulseset:master-SNAPSHOT"
     ))
-    exclude.set(setOf("**/generated/**")
+    filter {
+        exclude("**/generated/**")
+        include("**/kotlin/**")
+    }
```
JLLeitschuh: 👍

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
Tapchicoma: The problem is that `filterAction` is not available from configuration(returns default value), when task is created. Any clues how to get it on task creation time?

JLLeitschuh: Ooooh, interesting question. I'm thinking about it. Let me get back to you.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -83,6 +86,7 @@ open class KtlintCheckTask @Inject constructor(
 
     @TaskAction
     fun lint() {
+        logger.warn("Lint action")
```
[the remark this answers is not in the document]
Tapchicoma: Used for debugging and forget to remove :disappointed_relieved:

Tapchicoma:
@JLLeitschuh checked - works well. I think it is ready.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
```diff
@@ -6,21 +6,24 @@ import org.gradle.api.Project
 import org.gradle.api.tasks.util.PatternFilterable
 import org.gradle.util.GradleVersion
 
+typealias FilterApplier = (Action<PatternFilterable>) -> Unit
```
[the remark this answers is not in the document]
Tapchicoma: Fixed.

Issue #148, opened 2018-10-11
Use Junit 5


JLLeitschuh:
The only thing we have a dependnecy upon that will be kinda goofy for is the `TempDirRule`.
That can be fixed by doing this:
```kotlin
@ExtendWith(TempDirectory::class)
abstract class AbstractPluginTest {

    private lateinit var tempDir: Path

    @BeforeEach
    fun beforeEach(@TempDirectory.TempDir tempDir: Path) {
        this.tempDir = tempDir
    }

    val projectRoot: File
        get() = tempDir.toFile().resolve("plugin-test").apply { mkdirs() }
```

The `TmpDirectory` extension comes from
`testCompile(group = "org.junit-pioneer", name = "junit-pioneer", version = "0.2.2")`.

JLLeitschuh:
I don't really see much value add and it would require quite a rewrite of our tests.
I've been working with Junit 5 a lot and I like it. The `asserrtAll` support is really nice.

PR #154, opened 2018-10-26
WIP: fix additional source directories on android are not checked

### LINKED ISSUE #153: Alternate sourceSets are not detected
Plugin version: `6.2.0`
Ktlint version: `Bundled`
Project type: `Android`
Android Gradle plugin version: `3.4.0-alpha01` (happened in earlier 3.3.x ones too)

Config:

```groovy

ktlint {
    android = true
}
```

I have configured an alternative `sourceSet` for my kotlin sources:

```groovy
sourceSets {
    main {
        java.srcDirs += "src/main/kotlin"
    }
}
```

Once I do this, the plugin does not seem to be able to detect kotlin files in that directory. Kotlin files in `src/main/java` *are* detected.

Is this some misconfiguration on my end? I can create and share a sample project if needed.
Thanks for looking into this!

Tapchicoma:
Closes #153


Tapchicoma:
This fix only applies to android, though it is worth to add test for current tests setup.

Tapchicoma:
@JLLeitschuh for android I will try to address in #148 PR.

PR #157, opened 2018-11-01
New dev cycle

Tapchicoma:


plugin/buildSrc/src/main/kotlin/PluginDependencies.kt
```diff
@@ -1,6 +1,6 @@
 object PluginVersions {
-    val kotlin = "1.2.71"
-    val ktlintPlugin = "6.2.0"
+    val kotlin = "1.3.0"
```
JLLeitschuh: Just FYI, this is a newer version of Kotlin than the Gradle Kotlin DSL ships with. Just keep that in mind.

PR #158, opened 2018-11-01
Update Gradle publish plugin to 0.10.0 version.

Tapchicoma:


plugin/build.gradle.kts
```diff
@@ -107,11 +107,11 @@ pluginBundle {
     tags = listOf("ktlint", "kotlin", "linting")
 
     (plugins) {
-        register("ktlintPlugin") {
+        "ktlintPlugin" {
             id = "org.jlleitschuh.gradle.ktlint"
             displayName = "Ktlint Gradle Plugin"
         }
-        register("ktlintIdeaPlugin") {
+        "ktlintIdeaPlugin" {
```
JLLeitschuh: 😕 Are you sure this is right? `plugins` is a `NamedDomainObjectContainer`. Calling `register` is the way you're supposed to add elements to the collection. Does this actually work as is?

Tapchicoma: actually it complained on `register() {..}` after update.

Also check kotlin sample here: https://plugins.gradle.org/docs/publish-plugin - it uses approach that uses in this PR.

JLLeitschuh: Huh, you're right. I asked Eric about this in the Gradle Community Kotlin-DSL channel.
I think that the `gradlePlugin.plugins` block creating the elements in the `pluginBundle.plugins` `NamedDomainObjectCollection`.

PR #164, opened 2018-11-06
Added missing base plugin publishing info.

Tapchicoma:
Trying to fix following failed CI pipeline: https://travis-ci.org/JLLeitschuh/ktlint-gradle/builds/451544513

TL;DR base plugin is missing publishing id and display name.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -87,12 +87,6 @@ gradlePlugin {
             id = "org.jlleitschuh.gradle.ktlint"
             implementationClass = "org.jlleitschuh.gradle.ktlint.KtlintPlugin"
         }
-
-        register("ktlintBasePlugin") {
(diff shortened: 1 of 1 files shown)


plugin/build.gradle.kts
```diff
@@ -107,6 +107,10 @@ pluginBundle {
     tags = listOf("ktlint", "kotlin", "linting")
 
     (plugins) {
+        "ktlintBasePlugin" {
+            id = "org.jlleitschuh.gradle.ktlint-base"
+            displayName = "Ktlint Gradle Base Plugin"
+        }
```
JLLeitschuh: I was hoping that we could avoid publishing this as it's own independent artifact. But sure, if this is what we have to do.

Is there any way we can not do this?

Tapchicoma:
@JLLeitschuh removed all configuration for base plugin

PR #167, opened 2018-11-08
Use variants dynamically to put sourceSets

### LINKED ISSUE #153: Alternate sourceSets are not detected

tasomaniac:
Instead of resolving sourceSets from extension, resolve variants and use the sources defined in there.

Fixes #153

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -157,7 +157,7 @@ open class KtlintPlugin : Plugin<Project> {
                 target.extensions.configure(BaseExtension::class.java) { ext ->
                     ext.variants.all { variant ->
                         val filesCallable = Callable { variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories } }
-                        createTasks(variant.name, target.files(files))
+                        createTasks(variant.name, target.files(filesCallable))
                     }
                 }
             }
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -156,7 +156,7 @@ open class KtlintPlugin : Plugin<Project> {
             fun getPluginConfigureAction(): (Plugin<Any>) -> Unit = {
                 target.extensions.configure(BaseExtension::class.java) { ext ->
                     ext.variants.all { variant ->
-                        val files = variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories }
+                        val filesCallable = Callable { variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories } }
                         createTasks(variant.name, target.files(files))
                     }
                 }
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -17,6 +17,7 @@ import org.jetbrains.kotlin.gradle.plugin.experimental.KotlinNativeComponent
 import org.jetbrains.kotlin.gradle.plugin.konan.KonanArtifactContainer
 import org.jetbrains.kotlin.gradle.plugin.konan.KonanExtension
 import shadow.org.jetbrains.kotlin.gradle.plugin.tasks.KonanCompileTask
+import java.util.concurrent.Callable
 import kotlin.reflect.KClass
 
 /**
@@ -156,7 +157,9 @@ open class KtlintPlugin : Plugin<Project> {
             fun getPluginConfigureAction(): (Plugin<Any>) -> Unit = {
                 target.extensions.configure(BaseExtension::class.java) { ext ->
                     ext.variants.all { variant ->
-                        val filesCallable = Callable { variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories } }
+                        val filesCallable = Callable {
+                            variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories }
+                        }
                         createTasks(variant.name, target.files(filesCallable))
                     }
                 }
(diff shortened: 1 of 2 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -156,18 +155,10 @@ open class KtlintPlugin : Plugin<Project> {
...
-                            // only when task will need it.
-                            // Solves the problem of having additional source dirs in
-                            // current AndroidSourceSet, that are not available on eager
-                            // evaluation.
-                            createTasks(
-                                androidSourceSet.name,
-                                target.files(Callable { androidSourceSet.java.srcDirs })
-                            )
-                        }
+                    ext.variants.all { variant ->
```
tasomaniac: This `ext.variants` is coming from the `PluginUtils`. 

Here `all` makes it lazy loop over variants. When a variant is created, it is sourceSet is also set and contains all directories.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -156,18 +155,9 @@ open class KtlintPlugin : Plugin<Project> {
...
+                        val files = variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories }
```
JLLeitschuh: The problem with this is that it isn't lazily evaluated. 

You need to create this in a `Callable` for it to be lazily evaluated.

JLLeitschuh:
```suggestion
                        val filesCallable = Callable { variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories } }
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -156,18 +155,9 @@ open class KtlintPlugin : Plugin<Project> {
...
+                        createTasks(variant.name, target.files(files))
```
JLLeitschuh:
```suggestion
                        createTasks(variant.name, target.files(filesCallable))
```

JLLeitschuh:
Is there any good way to add a unit test for this?

tasomaniac:
Ktlint integration with a Kotlin project is tested already in the repo. The same should be done for an Android project too. We have some tests in Novoda plugin, I used those test to double check these changes. 

https://github.com/novoda/gradle-static-analysis-plugin/blob/develop/plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy

If a setup is done, I would be happy to contribute to tests.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
tasomaniac: This is actually not necessary thanks to `variants.all` above. There are sourceSet, flavor and buildType definitions in Android extension. At the point where a buildVariant is computed, all of them are already resolved.

JLLeitschuh: So a user can't add their own custom variant type?

JLLeitschuh:
@Tapchicoma Can unit tests against android be written or would that couple our test infrastructure to requiring the Android SDKs to be installed?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
tasomaniac: They can. `.all` is lazy and live loop over all variants. Default + custom variants arrive here together will additional custom sourceSets and sourceDirs inside them.

JLLeitschuh: Sure, you're looping over variants correctly, the problem you have is that what if you end up with user code in a scenario like this:

```kotlin
fun getPluginConfigureAction(): (Plugin<Any>) -> Unit = {
    target.extensions.configure(BaseExtension::class.java) { ext ->
        ext.variants.all { variant ->
            val files = variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories }
            createTasks(variant.name, target.files(files))

            // Somewhere off in user build code land later on...
            variant.addJavaSourceFoldersToModel(File(......))
        }
    }
}
```

Now, because of the eager evaluation of `variant.sourceSets.flatMap { sourceSet -> sourceSet.javaDirectories }` the file that is added later is completely missed.

tasomaniac: Hmmm, this is interesting. I will try this out to see what happens. The recommended way of adding source dirs is to use sourceSets extension on Android extension but we should change this still to be safe than sorry. 👍

Tapchicoma:
@tasomaniac thank you for your contribution!

As we discussed in #153 and #170 - can you update your PR:
Instead of creating check task for all android variant sources, create a meta task per variant (adding `Variant` for a task name - for example, `ktlintKellerbierDebugVariantCheck`) and include a dependency on specific source set task. As far, as I see from your PR, it will be possible to do, just need to extract source set check and format tasks name creation based on `SourceSet` name and use it, when declaring dependency on check (format) task.

tasomaniac:
Hey, I believe that would be the best. But I need to start from scratch to do that. I will just close this so anybody can take it.

Tapchicoma:
@tasomaniac I think, I will approach this after #180 will be merged :+1:

tasomaniac:
Thanks @Tapchicoma

PR #173, opened 2018-11-11
Added a faq entry how to use ReporterType in external Gradle scripts.

### LINKED ISSUE #165: ClassNotFoundException for ReporterType
Hi, 

With the recent changes, the string definition of reporters list was replaced with an enum type. I tried many many times but I'm getting ClassNotFoundException. The import is also not recognized. 

Would it be possible to change `reporters` from being a list to a data structure. The structure most common with the standard code analysis tools that come with Gradle would be awesome. They use something like

```
reporters {
    xml.enabled = true
}
```

What I'm doing different in my project that I have a separate `static-analysis.gradle` file where I apply ktlint. And then I put this into each subproject. I don't directly configure ktlint directly in `build.gradle` files. 

Maybe that is why I keep having problems. 

I will try to provide an example project.

Tapchicoma:
Closes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,16 @@ No. This approaches are not equivalent how they work. The problem that
 plugin may not find some of kotlin plugins if both approaches are used
 in the project configuration. Especially it is related to Android plugin.
 
+- How can I import `ReporterType` in my external Gradle script?
+
+Gradle doesn't allow to have import statements in external gradle script files (for example: `quality/foo.gradle`).
+To solve this - add following line in your root `build.gradle` file:
+
+``` gradle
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -238,6 +238,16 @@ No. This approaches are not equivalent how they work. The problem that
 plugin may not find some of kotlin plugins if both approaches are used
 in the project configuration. Especially it is related to Android plugin.
 
+- How can I import `ReporterType` in my external Gradle script?
+
+Gradle doesn't allow to have import statements in external gradle script files (for example: `quality/foo.gradle`).
```
JLLeitschuh: What?! This sounds like a bug in Gradle. Care to open or link to an existing bug here?

Issue #182, opened 2018-12-09
Add muliplatform targets meta tasks


Tapchicoma:
Currently there is no way to check sources only for specific target in new kotlin multiplatform plugin (for example, only for common or only for js).

We should add related tasks that will trigger actual tasks for all source sets that are in the target.

Open question: should such tasks also check sources from dependent targets?

tasomaniac:
This is because of the changes in the new version, right? They made it very flexible, is that the reason?

Tapchicoma:
Current main problem is that in mpp extension targets and related sourcesets are detached from each other. So it is very not straightforward to get map sourcesets to target. And not possible to get source sets that was added as dependency using `dependOn()`.

I think that this issue is kind of related: https://youtrack.jetbrains.com/issue/KT-26962, but want to clarify it.

Tapchicoma:
Directly related issue: https://youtrack.jetbrains.com/issue/KT-28749

JLLeitschuh:
Is this a use case we really want to cover?

If so, do we want to allow the user to opt-out of certain tasks being generated?
Can we do so with the `KtlintExtension` in such a way that enables this filtering?

@Tapchicoma out of curiosity, do we currently automatically attach the ktlintCheck tasks to the `check` task?

I'd also like to avoid the situation where we are creating way too many marker tasks which will greatly expand the number of tasks we have to ensure we don't break in the future (eg. accidental or intentional renaming).

Tapchicoma:
@JLLeitschuh yes, running `check` task will run `ktlintCheck` task that will trigger run check for each source set in the current project.

> If so, do we want to allow the user to opt-out of certain tasks being generated?
> Can we do so with the KtlintExtension in such a way that enables this filtering?

Yes, we can add such config. Boolean property, imho, will be enough.

> I'd also like to avoid the situation where we are creating way too many marker tasks which will greatly expand the number of tasks we have to ensure we don't break in the future (eg. accidental or intentional renaming).

I think this issue is valid as before new MPP plugin each target was in it own gradle module, so check for target can be run separately.

PR #183, opened 2018-12-09
Add meta tasks for Android variants

### LINKED ISSUE #170: Add android variants check and format tasks
Currently, plugin checks SourceSet for particular android variant, but it doesn't check sources that variant depends on as well.

Need to add special meta tasks for each android variant that will trigger tasks for all dependent variant sources.

Tapchicoma:
This PR adds android variants meta tasks, that trigger tasks for all source sets in given variant.

For example running `ktlintKellerbierDebugCheck` meta task triggers following tasks:
```bash
$ ./gradlew :samples:android-app:ktlintKellerbierDebugCheck -m

:samples:android-app:ktlintDebugSourceSetCheck SKIPPED
:samples:android-app:ktlintKellerbierDebugSourceSetCheck SKIPPED
:samples:android-app:ktlintKellerbierSourceSetCheck SKIPPED
:samples:android-app:ktlintMainSourceSetCheck SKIPPED
:samples:android-app:ktlintKellerbierDebugCheck SKIPPED
```

Also added limited support for android projects in new Kotlin multiplatform plugin:
```bash
$ ./gradlew -m :samples:kotlin-mpp-android:ktlintKellerbierDebugAndroidLibCheck

:samples:kotlin-mpp-android:ktlintAndroidLibDebugSourceSetCheck SKIPPED
:samples:kotlin-mpp-android:ktlintAndroidLibKellerbierDebugSourceSetCheck SKIPPED
:samples:kotlin-mpp-android:ktlintAndroidLibKellerbierSourceSetCheck SKIPPED
:samples:kotlin-mpp-android:ktlintAndroidLibMainSourceSetCheck SKIPPED
:samples:kotlin-mpp-android:ktlintKellerbierDebugAndroidLibCheck SKIPPED
```

Currently variant task name is inconsistent with source sets tasks names, but that will be addressed in #182 issue.

Closes #170.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-  - ?
+  - Meta tasks to run check or format on all sources in android variant. (#170)
+  Example: In an Android project with `foo` flavor,
+  `ktlintFooDebugSourceSetCheck` task will check the `foo` sourceSet (not main).
+  `ktlintFooDebugCheck` meta task will check all the sourceSets for `fooDebug` build variant.
 ### Changed
   - Update Kotlin to `1.3.10` version
   - Breaking: check/format tasks for specific source sets
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -396,6 +396,7 @@ open class KtlintPlugin : Plugin<Project> {
     /*
      * Helper functions used until Gradle Script Kotlin solidifies it's plugin API.
      */
+
     private inline fun <reified T : Any> Project.theHelper() =
         theHelper(T::class)
 
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -365,7 +365,7 @@ open class KtlintPlugin : Plugin<Project> {
         multiplatformTargetName: String? = null
     ): Task {
         val taskName = "ktlint${variantName.capitalize()}${multiplatformTargetName?.capitalize() ?: ""}Check"
-        return tasks.findByName(taskName) ?: task(taskName).apply {
+        return tasks.maybeCreate(taskName).apply {
             group = VERIFICATION_GROUP
             description = "Runs ktlint on all kotlin sources for android $variantName variant in this project."
         }
@@ -382,7 +382,7 @@ open class KtlintPlugin : Plugin<Project> {
         multiplatformTargetName: String? = null
     ): Task {
         val taskName = "ktlint${variantName.capitalize()}${multiplatformTargetName?.capitalize() ?: ""}Format"
-        return tasks.findByName(taskName) ?: task(taskName).apply {
+        return tasks.maybeCreate(taskName).apply {
             group = FORMATTING_GROUP
             description = "Runs ktlint formatter on all kotlin sources for android $variantName" +
                 " variant in this project."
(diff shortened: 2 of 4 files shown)


Tapchicoma:
@tasomaniac may you also take a look on this PR?

CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-  - ?
+  - Meta tasks to run check or format on all sources in android variant. (#170)
```
tasomaniac: Nice 👌

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -324,20 +360,42 @@ open class KtlintPlugin : Plugin<Project> {
             description = "Runs ktlint on all kotlin sources in this project."
         }
 
+    private fun Project.getAndroidVariantMetaKtlintCheckTask(
+        variantName: String,
+        multiplatformTargetName: String? = null
+    ): Task {
+        val taskName = "ktlint${variantName.capitalize()}${multiplatformTargetName?.capitalize() ?: ""}Check"
+        return tasks.findByName(taskName) ?: task(taskName).apply {
```
tasomaniac: Instead can you use `tasks.maybeCreate()` that conditionally creates a task and returns the already available task if so.

JLLeitschuh: Unfortunately, all of these options are eagerly evaluated. I'd love if there was a way to lazily instantiate this task.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -324,20 +360,42 @@ open class KtlintPlugin : Plugin<Project> {
...
+    }
+
     private fun setCheckTaskDependsOnKtlintCheckTask(project: Project, ktlintCheck: Task) {
         project.tasks.findByName("check")?.dependsOn(ktlintCheck)
     }
 
     /*
      * Helper functions used until Gradle Script Kotlin solidifies it's plugin API.
      */
-
```
JLLeitschuh: I think this is a line of documentation for all lower lines, so this newline gap was intentional.

JLLeitschuh:
One concern I want to bring up is the potential for accidental task name collisions.
Are there any cases that anyone can think of whereby declaring source sets in certain ways you could unintentionally get task name declaration collisions?

If there are, please figure out how to prevent these task name collisions.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
tasomaniac: I'm not sure actually. findByName is also not lazy. `maybeCreate` looks like a drop-in replacement of what you do here. 

At the end it is not crucial. Either is fine.

tasomaniac:
I don't see any problems. It will bring back the old functionality of the tasks back.

The breaking change would be that the type of the tasks has changed together with the report file destinations. We are using that actually but I will make the necessary changes to support the new structure. I guess the `meta task` wording in the changelog already captures this.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: I think, this can addressed in next PR with task configuration avoidance. For now, I changed it to `maybeCreate()`

Tapchicoma:
> One concern I want to bring up is the potential for accidental task name collisions.
Are there any cases that anyone can think of whereby declaring source sets in certain ways you could unintentionally get task name declaration collisions?

I agree with @tasomaniac that there is no problem with this.

JLLeitschuh:
Okay, if you are both unconcerned, then go for it if this is ready to go.

tasomaniac:
Nice. Are you guys going to release a new version now or does it need more stuff in?

Tapchicoma:
@tasomaniac As next release will anyway break api - I also want to bump minimal supported Gradle version to `5.0.0` and add task configuration avoidance. Then we can do new `7.0.0` release.

tasomaniac:
Hey about the task configuration avoidance: there is a way to make it available in a backward compatible way. I found this example from official docs: https://github.com/melix/jmh-gradle-plugin/blob/a034aa88805b7a06fa9c5a825d573554b2aa23e2/src/main/groovy/me/champeau/gradle/JMHPlugin.groovy#L289-L296

Especially with Kotlin extension functions, this can be done in a very nice way.

Issue #185, opened 2018-12-16
Use Gradle 5.0


Tapchicoma:
For now blocked by https://github.com/JetBrains/kotlin-native/issues/2440

JLLeitschuh:
So we can't update to Gradle 5.0 internally currently? Can we run some of the tests against Gradle 5.0?

Tapchicoma:
We can update plugin project to Gradle `5.0`, but leave samples on `4.10`.

Tapchicoma:
Fixed via #235

PR #186, opened 2018-12-16
Task configuration avoidance

### LINKED ISSUE #174: Cannot apply plugin into Android multi module project
I cannot find a way to apply this plugin into my Android project. I tried both approaches:

```
plugins {
    id "org.jlleitschuh.gradle.ktlint" version "6.3.0"
}
```
and 

```
buildscript {
  repositories {
    maven {
      url "https://plugins.gradle.org/m2/"
    }
  }
  dependencies {
    classpath "gradle.plugin.org.jlleitschuh.gradle:ktlint-gradle:6.3.0"
  }
}

```

Also I tried to apply it in multiple gradle files (root file and module files). I keep getting following error:

```
Unable to load class 'org.gradle.api.provider.SetProperty'.
Possible causes for this unexpected error include:<ul><li>Gradle's dependency cache may be corrupt (this sometimes occurs after a network connection timeout.)
Re-download dependencies and sync project (requires network)</li><li>The state of a Gradle build process (daemon) may be corrupt. Stopping all Gradle daemons may solve this problem.
Stop Gradle build processes (requires restart)</li><li>Your project may be using a third-party plugin which is not compatible with the other plugins in the project or the version of Gradle requested by the project.</li></ul>In the case of corrupt Gradle processes, you can also try closing the IDE and then killing all Java processes.
```

Any of solutions proposed by error didn't help in any way.

Gradle wrapper version is 4.4 and gradle plugin version is 3.0.1.
I am willing to provide any further required information.

Tapchicoma:
This PR adds plugin tasks configuration avoidance: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

For example, running `./gradlew :samples:kotlin-ks:help -Dorg.gradle.internal.tasks.stats` produces following info for the project:
```
Task counts: Old API 3, New API 32, total 35
Task counts: created 17, avoided 18, %-lazy 52

Task types that were created with the old API
class org.jetbrains.kotlin.gradle.tasks.KotlinCompile 2
class org.jetbrains.kotlin.gradle.tasks.InspectClassesForMultiModuleIC 1

Task types that were registered with the new API but were created anyways
class org.gradle.api.DefaultTask 5
class org.gradle.api.tasks.compile.JavaCompile 2
class org.gradle.language.jvm.tasks.ProcessResources 2
class org.gradle.api.tasks.compile.GroovyCompile 2
class org.gradle.api.tasks.testing.Test 1
class org.gradle.api.tasks.Delete 1
class org.gradle.api.tasks.bundling.Jar 1
Task counts: Old API 3, New API 32, total 35
Task counts: created 17, avoided 18, %-lazy 52

Task types that were created with the old API
class org.jetbrains.kotlin.gradle.tasks.KotlinCompile 2
class org.jetbrains.kotlin.gradle.tasks.InspectClassesForMultiModuleIC 1

Task types that were registered with the new API but were created anyways
class org.gradle.api.DefaultTask 5
class org.gradle.api.tasks.compile.JavaCompile 2
class org.gradle.language.jvm.tasks.ProcessResources 2
class org.gradle.api.tasks.compile.GroovyCompile 2
class org.gradle.api.tasks.testing.Test 1
class org.gradle.api.tasks.Delete 1
class org.gradle.api.tasks.bundling.Jar 1
```
Note, that this is not all output, as this param prints such info for all projects (composite as well).

As task configuration avoidance was introduces in Gradle `4.9` - set minimal Gradle supported version to `4.10` (closes #174). Setting minimal supported Gradle version to `5.0` is not possible for now, see #185 for details.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -354,43 +362,64 @@ open class KtlintPlugin : Plugin<Project> {
...
-        project.tasks.findByName("check")?.dependsOn(ktlintCheck)
+    private fun setCheckTaskDependsOnKtlintCheckTask(
+        project: Project,
+        ktlintCheck: TaskProvider<KtlintCheckTask>
+    ) {
+        try {
+            project.tasks.named("check").configure { it.dependsOn(ktlintCheck) }
+        } catch (e: UnknownDomainObjectException) {
+            project.logger.info("Check task is not available.")
+        }
```
JLLeitschuh: @big-guy Is there a better way of doing this in Gradle 5.0?

@Tapchicoma I don't know if the logging is valuable here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -31,8 +32,11 @@ internal fun createConfiguration(target: Project, extension: KtlintExtension) =
         )
     }
 
-internal inline fun <reified T : Task> Project.taskHelper(name: String, noinline configuration: T.() -> Unit): T {
-    return this.tasks.create(name, T::class.java, configuration)
+internal inline fun <reified T : Task> Project.taskHelper(
+    name: String,
+    noinline configuration: T.() -> Unit
+): TaskProvider<T> {
+    return this.tasks.register(name, T::class.java, configuration)
```
JLLeitschuh: 👍

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -354,43 +362,64 @@ open class KtlintPlugin : Plugin<Project> {
...
-    private fun Project.getMetaKtlintCheckTask(): Task = this.tasks.findByName(CHECK_PARENT_TASK_NAME)
-        ?: this.task(CHECK_PARENT_TASK_NAME).apply {
+    private fun Project.getMetaKtlintCheckTask(): TaskProvider<Task> = try {
+        this.tasks.named(CHECK_PARENT_TASK_NAME)
+    } catch (ignored: UnknownDomainObjectException) {
+        taskHelper(CHECK_PARENT_TASK_NAME) {
             group = VERIFICATION_GROUP
             description = "Runs ktlint on all kotlin sources in this project."
         }
+    }
```
JLLeitschuh: What's going on here? I don't understand.

JLLeitschuh:
If there's a cheaper way to do this than catching exceptions, that would be awesome.

Tapchicoma:
> If there's a cheaper way to do this than catching exceptions, that would be awesome.

Was my thought as well. Tough I haven't found anything that avoids task creation, but allows to configure it :disappointed:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: I try to get lazy task and if it fails to find - create a new one.

JLLeitschuh: Is there a way to do this without an exception being caught as that's expensive?
Also, if it can't be simplified, can you generify this bit of `try` & `catch` logic to be a generic method?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
JLLeitschuh: Instead of doing this, only configure the `check` task when the `base` plugin is applied.

tasomaniac: It is not only about checking if `check` task is available or not. But also about not causing an early configuration. If you configure it without `named`, then you cause other configurations to happen too.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -31,8 +32,11 @@ internal fun createConfiguration(target: Project, extension: KtlintExtension) =
         )
     }
 
-internal inline fun <reified T : Task> Project.taskHelper(name: String, noinline configuration: T.() -> Unit): T {
-    return this.tasks.create(name, T::class.java, configuration)
+internal inline fun <reified T : Task> Project.taskHelper(
```
tasomaniac: Can you call this `register`? Since the params are not the same, using `register` is still possible. It would be more intuitive than `taskHelper`. What do you think?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -354,43 +362,64 @@ open class KtlintPlugin : Plugin<Project> {
...
 
     private fun Project.getAndroidVariantMetaKtlintCheckTask(
         variantName: String,
         multiplatformTargetName: String? = null
-    ): Task {
+    ): TaskProvider<Task> {
         val taskName = "ktlint${variantName.capitalize()}${multiplatformTargetName?.capitalize() ?: ""}Check"
-        return tasks.maybeCreate(taskName).apply {
```
tasomaniac: I see. Maybe create is not available anymore, and they didn't provide any lazy alternative, it seems.

Tapchicoma:
@JLLeitschuh rethink meta tasks creation and configuration. Main idea - meta tasks added immediately if sources sets are not empty. This allow to avoid `try {} catch (e: UnknownDomainObjectException) {}` approach.

/cc @tasomaniac

tasomaniac:
`main` sourceSet is added to every build flavor in Android. So meta tasks will never be empty. I think this makes a lot of sense.

tasomaniac:
LGTM. Much cleaner.

tasomaniac:
Created this on Gradle for maybeRegister https://github.com/gradle/gradle/issues/8106

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -11,12 +11,15 @@ import org.gradle.api.internal.HasConvention
...
 import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.plugin.experimental.KotlinNativeComponent
 import org.jetbrains.kotlin.gradle.plugin.konan.KonanArtifactContainer
 import org.jetbrains.kotlin.gradle.plugin.konan.KonanExtension
+import org.jetbrains.kotlin.utils.addToStdlib.ifNotEmpty
```
JLLeitschuh: What is the `addToStdlib` package?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -57,6 +60,11 @@ open class KtlintPlugin : Plugin<Project> {
         val ktLintConfig = createConfiguration(target, extension)
         val multiplatformExtension = target.extensions.getByType(KotlinMultiplatformExtension::class.java)
 
+        multiplatformExtension.sourceSets.ifNotEmpty {
+            target.createMetaKtlintCheckTask()
+            target.createMetaKtlintFormatTask()
+        }
```
JLLeitschuh: What if the source set is empty when this plugin is evaluated but has sources added to it by a later plugin?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -247,6 +276,11 @@ open class KtlintPlugin : Plugin<Project> {
         return {
             val ktLintConfig = createConfiguration(project, extension)
 
+            project.components.withType(KotlinNativeComponent::class.java).ifNotEmpty {
```
JLLeitschuh: Remind me the difference between `components` and `extensions`?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: Interesting, I don't know about `addToStdlib` package :thinking:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: `Component` is configuration for some compile/output artifact, see https://docs.gradle.org/current/javadoc/org/gradle/api/component/SoftwareComponent.html

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -406,4 +386,24 @@ open class KtlintPlugin : Plugin<Project> {
     private inline fun <reified T> Convention.getPluginHelper() = getPlugin(T::class.java)
 
     private fun Project.isConsolePlain() = gradle.startParameter.consoleOutput == ConsoleOutput.Plain
+
+    private inner class PluginHolder(
```
JLLeitschuh: I don't think this needs to be an `inner` class. Using the `inner` is basically the opposite of using `static` on a nested class in Java.

Unless you are scoping this class over something I'm not seeing, I think you can remove it.

JLLeitschuh:
One final change and this will be good to go.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: Removed `inner` and force-pushed.

PR #188, opened 2019-01-07
add ktlintFileFormat task that only formats specified files

gabrielittner:
This allows passing a list of files that ktlint should format to a newly created Gradle task. Our use case is to run ktlint as pre-commit hook, however just running it over the complete code base takes too long. This allows to only run it over files that actually changed. 

Execution times when running over a code base with no changes:
```
./gradlew help - 2s
./gradlew ktlintFileFormat - 3s
./gradlew ktlintFormat - 42s
``` 

I had to add a new task for this because Gradle's `@Option` requires directly invoking the task. So adding it to the normal check/format tasks wouldn't work because those are invoked by the meta tasks. An alternative to this would be using something else like properties and then filter directly in `KtlintCheckTask` when that property is set.


samples/android-app/build.gradle.kts
```diff
@@ -44,7 +44,7 @@ configure<AppExtension> {
 }
 
 dependencies {
-    "implementation"("org.jetbrains.kotlin:kotlin-stdlib-jre7:${SamplesVersions.kotlin}")
+    "implementation"("org.jetbrains.kotlin:kotlin-stdlib-jdk7:${SamplesVersions.kotlin}")
```
gabrielittner: `kotlin-stdlib-jre7` doesn't exist anymore in 1.3.10.

JLLeitschuh:
Wouldn't it be better to modify our task so that it only runs on files that are out of date?

https://docs.gradle.org/current/dsl/org.gradle.api.tasks.incremental.IncrementalTaskInputs.html

gabrielittner:
That only helps if you recently ran ktlintFormat on your machine and is also affected by things like branch changes.

JLLeitschuh:
If you are using the Gradle Build Cache, I think it will ignore things like branch changes and how recently you ran the `ktlintFormat` task.

gabrielittner:
Enabling build cache for the format task only helps for modules where nothing changed at all. Otherwise the cache key doesn't match and you still have to format the whole module. Not saying that making `ktlintFormat` incremental and adding build cache support wouldn't help, but there still would be many situations where it takes more time than just passing in the files that we want to format. 

An alternative would be allowing to set a `(FileTree) -> FileTree` through the extension. That way I could just do the filtering in our build files. `PatterFilterable` unfortunately isn't enough to filter for exact files.

JLLeitschuh:
> That way I could just do the filtering in our build files. PatterFilterable unfortunately isn't enough to filter for exact files.

I don't understand. Why isn't it enough?

gabrielittner:
If you do `include("**/MyClass.kt")` the FileTree will include `MyClass`, if you do `include("**/module/src/main/kotlin/com/example/MyClass.kt")` it won't include the class.

JLLeitschuh:
Okay, these are my current thoughts on this PR.

First off, I want to thank you for making a contribution to this project. I'm glad that it's working well for you.

The thing about this PR that I'm concerned about is that it adds additional bloat to the plugin. I'm not against adding it if there's a valid use case, but I want to make sure that the problem can't be solved with different solutions.

> If you do `include("**/MyClass.kt")` the FileTree will include `MyClass`, if you do `include("**/module/src/main/kotlin/com/example/MyClass.kt")` it won't include the class.

This sounds like a bug. Is it not?

Doesn't [PatternFilterable include​(Spec<FileTreeElement> includeSpec)](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html#include(org.gradle.api.specs.Spec)) cover your use case?
It might require some pasring of the input in your build script file, but I think it would work for you, would it not?

If this doesn't work, you can get access to the `FileTree` using the [SourceTask::getSource](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceTask.html#getSource()) method on the `KtlintFormat` task.

In your build file, just do something like:
```kotlin
tasks.withType<KtlintFormatTask>().configureEach {
     source.... // This is of type FileTree
}
```

Does this cover your use case or is there something I'm still missing?

PR #190, opened 2019-01-09
Avoid error due to failing resolution of aapt2-proto

devisnik:
When building the plugin I get an error due to a failed resolution of `aapt2-proto.jar`.

screenshot 2019-01-09 at 11 53 13

This is an known issue, see https://stackoverflow.com/questions/52944598/could-not-find-aapt2-proto-jar


JLLeitschuh:
@Tapchicoma Thoughts? This is interesting. I didn't realize Gradle cared about the order of repos.

PR #195, opened 2019-01-30
Fix wrong up-to-date state of format task.

### LINKED ISSUE #194: ktlintFormat is up-to-date when it shouldn't be
To reproduce:
1. Do a small change that would be fixed by ktlint (e.g. inserting a space before a `,`) 
2. Run `gw ktlintFormat` 
3. The change should be undone now
4. Do the same change you did in 1 again
5. Run `gw ktlintFormat` again
-> The task will be up-to-date and the change still be there

Tapchicoma:
Fixes #194: Format task may produce up-to-date state, if sources was restored to pre-format state, so task input considered to be cached from success run.

Current solution is not perfect, but, currently, there are no better ways in Gradle:
- on first run format changes the sources and produce stable output content 
- on second run gradle produces stable input content with the same output
- on third run task is considered "UP-TO-DATE" and cached

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintFormatTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintFormatTask.kt
@@ -1,13 +1,31 @@
 package org.jlleitschuh.gradle.ktlint
 
+import org.gradle.api.file.FileTree
 import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.OutputFiles
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.SkipWhenEmpty
 import org.gradle.process.JavaExecSpec
 import javax.inject.Inject
 
+@CacheableTask
 open class KtlintFormatTask @Inject constructor(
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintFormatTask.kt
```diff
@@ -1,13 +1,31 @@
...
     override fun additionalConfig(): (JavaExecSpec) -> Unit = {
         it.args("-F")
     }
+
+    /**
+     * Fixes the issue when input file is restored to pre-format state and running format task again fails
+     * with "up-to-date" task state.
+     *
+     * Note this approach sets task to "up-to-date" only on 3rd run when both input and output sources are the same.
+     */
```
JLLeitschuh: @nedtwigg I know you fixed this same problem with Spotless. Care to offer any insights into how you resolved it?

nedtwigg: I wrote it up [here](https://discuss.gradle.org/t/two-cases-where-the-incremental-api-is-lacking/24266), and included links to the PR's that fixed it.

Tapchicoma: It is kind of complex solution, I'd prefer not introduce it here.

PR #196, opened 2019-01-31
Set minimal supported ktlint version to 0.22.0.

Tapchicoma:
This should simplify code and tests.


JLLeitschuh:
If you really don't want to continue to maintain earlier version support, I'm fine with this change. But personally, I don't have any problem with supporting early versions of ktlint.

Tapchicoma:
I don't think that somebody still uses older versions of ktlint, but upgrades this plugin - doesn't make sense to me. Also if there is such person - they still can use older version of this plugin.

PR #202, opened 2019-02-04
Notify usage of vulnerable ktlint version.

Tapchicoma:


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -86,6 +86,7 @@ open class KtlintCheckTask @Inject constructor(
     @TaskAction
     fun lint() {
         checkMinimalSupportedKtlintVersion()
+        checkCWEKtlintVersion()
```
JLLeitschuh: This is a vulnerability in both the `check` and the `lint` tasks. I want to make sure this is applied to both plugins.

Tapchicoma: Yes, as format extends check task - it works for both.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -126,6 +127,17 @@ open class KtlintCheckTask @Inject constructor(
         }
     }
 
+    private fun checkCWEKtlintVersion() {
+        if (!ruleSets.get().isNullOrEmpty() &&
+            SemVer.parse(ktlintVersion.get()) < SemVer(0, 30, 0)) {
```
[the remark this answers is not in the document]
JLLeitschuh: Lord. haha. If I'd just looked one line above. My mistake.

PR #207, opened 2019-02-21
Simplify tests setup.

### LINKED ISSUE #150: Simplify tests using GradleRunner.withPluginClassPath()
We could significantly simplifies the tests by using the [GradleRunner.withPluginClassPath()](https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html#withPluginClasspath--) API as proposed in this github issue comment:
https://github.com/gradle/kotlin-dsl/issues/1178#issuecomment-429861533

It would let us use the `plugins` block in our unit tests instead of using the `buildScript` and then the `apply`.

This will also allow us to unit test with the Gradle Kotlin DSL. The `ktlint` accessors would automatically get generated by gradle and we could make sure that the API was what we wanted in our unit tests.

Tapchicoma:
Now tests are using `GradleRunner.withPluginClasspath()` method to inject plugin under test.

Closes #150.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,17 @@ Building the plugin: `./plugin/gradlew build`
 
 On how to run the current plugin snapshot check on sample projects: `./gradlew ktlintCheck`
 
+### Running tests from [IDEA IDE](https://www.jetbrains.com/idea/)
+
+To run tests in [IDEA IDE](https://www.jetbrains.com/idea/), 
+firstly you need to run following gradle task (or after any dependency change):
+
+```bash
+$ ./plugin/gradlew pluginUnderTestMetadata
+```
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -276,6 +276,17 @@ Building the plugin: `./plugin/gradlew build`
...
+
+To run tests in [IDEA IDE](https://www.jetbrains.com/idea/), 
+firstly you need to run following gradle task (or after any dependency change):
+
+```bash
+$ ./plugin/gradlew pluginUnderTestMetadata
+```
+
+Optionally you can add this step test run configuration.
+
```
JLLeitschuh: 👍 Thanks!

Issue #210, opened 2019-03-04
Add support for editorconfig path


philippeagra:
Hello,
it would be cool if we could configure the path of the `.editorconfig` file (using ktlint `--editorconfig` flag) within KtlintExtension.

Tapchicoma:
You want to provide `.editorconfig` file that is placed outside project?

philippeagra:
Indeed, I wanted to use an `.editorconfig` in the build folder (generated by an other plugin) to avoid having the same `.editorconfig` in all my projects. I know i can use a ruleset but i have to implement all the standard rule to avoid using the `.editorconfig` file.

PR #211, opened 2019-03-05
Fix .editorconfig in project root is not considered as tasks input.

### LINKED ISSUE #209: .editorconfig change seems to be ignored
I am not sure if this is Ktlintgradle bug or ktlint itself.

I have a mono repo with multiple projects. `.editirconfig` exists in the root of mono repo, but The Android project does not have it's own `.editirconfig` file:

Monorepo root
- `.editirconfig`
    - `android`
    - `iOs`

it looks like changes to `.editirconfig` are ignored (due to cache?). 

Steps 
1. Run ktlint gradle inside Android project `./gradlew ktlintcheck` that should pass
2. Make change in the `.editirconfig` that should make ktlint check fail eg. add `indent_size = 2`
3. Run ktlint 

Result
All checks are OK

Expected results
Some checks should fail

I discovered that by editing one of the files in the Android project (let's say added unwanted space) the check will start failing, so this leads to a conclusion that maybe cache is not invalidated when the `.editirconfig` file is updated.

Tapchicoma:
This prevents task to reset UP-TO-DATE state on any changes to this file.

Fixes #209

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -39,7 +39,10 @@ open class KtlintCheckTask @Inject constructor(
 
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:InputFiles
-    internal val editorConfigFiles: FileCollection = getEditorConfigFiles(project)
+    internal val editorConfigFiles: FileCollection by lazy(LazyThreadSafetyMode.NONE) {
+        // Gradle will lazy evaluate this task input only on task execution
+        getEditorConfigFiles(project)
+    }
 
     @get:Input
     internal val ktlintVersion: Property<String> = objectFactory.property()
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
@@ -57,7 +57,7 @@ private tailrec fun searchEditorConfigFiles(
     fileCollection: FileCollection
 ): FileCollection {
     val editorConfigFC = projectPath.resolve(EDITOR_CONFIG_FILE_NAME)
-    val outputCollection = if (editorConfigFC != null) {
+    val outputCollection = if (editorConfigFC.toFile().exists()) {
         fileCollection.plus(project.files(editorConfigFC))
     } else {
         fileCollection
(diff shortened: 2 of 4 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -57,15 +57,15 @@ private tailrec fun searchEditorConfigFiles(
     fileCollection: FileCollection
 ): FileCollection {
     val editorConfigFC = projectPath.resolve(EDITOR_CONFIG_FILE_NAME)
-    val outputCollection = if (editorConfigFC != null) {
+    val outputCollection = if (editorConfigFC.toFile().exists()) {
```
JLLeitschuh: Is this evaluated eagerly or at task execution time?

Tapchicoma: I think eagerly on task configuration phase

JLLeitschuh: Is that a problem? Do we care if the file is added/modified by a previous task?

Tapchicoma: I would say - we care, see, for example, #210 discussion.

What is the best way to do it? Current implementation: https://github.com/JLLeitschuh/ktlint-gradle/blob/0b019e0c0d99c2a34c2ff40fef61a1fefc21eebf/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt#L42
 
Use `by lazy { .. }` at this line or use `ObjectFactory.property { .. }`?

JLLeitschuh: Make it an input to the task that's a `FileProperty`? Honestly, I'd have to dig into where this code is actually called and see why you're adding it to this particular `FileCollection` to understand better what's going on.

Tapchicoma: I think using `by lazy { ... }` will work here.

Tapchicoma: Tested, works good - `getEditorconfigFiles()` is only evaluated on task execution when Gradle gets inputs for the task.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -39,7 +39,9 @@ open class KtlintCheckTask @Inject constructor(
 
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:InputFiles
-    internal val editorConfigFiles: FileCollection = getEditorConfigFiles(project)
+    internal val editorConfigFiles: FileCollection by lazy(LazyThreadSafetyMode.NONE) {
+        getEditorConfigFiles(project)
+    }
```
JLLeitschuh: Can you leave a comment explaining why this is intentionally `lazy`? Other than that, this PR looks good!

Tapchicoma: Added comment and force pushed.

PR #212, opened 2019-03-07
Add task for kotlin script files

### LINKED ISSUE #98: Add a task to check project kotlin dsl files
Currently, while `KtlintCheck` task supports checking `kts` files, plugin itself doesn't add any task for this files.

Plugin should add task that checks project kotlin gradle dsl files.

Tapchicoma:
Fixes #98 

@JLLeitschuh can you check if I am doing right with extension and additional paths for kotlin script files?

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
@@ -3,10 +3,13 @@ package org.jlleitschuh.gradle.ktlint
 import org.gradle.api.Action
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.file.ConfigurableFileTree
 import org.gradle.api.tasks.util.PatternFilterable
 import org.gradle.util.GradleVersion
 
 internal typealias FilterApplier = (Action<PatternFilterable>) -> Unit
+internal typealias KotlinScriptAdditionalPathApplier = (ConfigurableFileTree) -> Unit
 
 /**
  * The base Ktlint plugin that all other plugins are built on.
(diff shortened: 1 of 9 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
```diff
@@ -24,11 +27,27 @@ open class KtlintBasePlugin : Plugin<Project> {
             }
         }
 
+        val kotlinScriptAdditionalPathApplier = { fileCollection: FileCollection ->
```
JLLeitschuh: Do you want to be explicit about the type here, using `KotlinScriptAdditionalPathApplier` just for sanities sake? If you don't think it adds any value, feel free to not add it.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
```diff
@@ -24,11 +27,27 @@ open class KtlintBasePlugin : Plugin<Project> {
...
+            val configureAction = Action<Task> {
+                with(it as KtlintCheckTask) {
+                    fileCollection.forEach {
+                        val additionalPath = project.fileTree(it)
+                        additionalPath.include("*.kts")
+                        source = source.plus(additionalPath)
+                    }
```
JLLeitschuh: Do you really need to iterate over the entire `fileCollection` here? I feel like that will just slow down the configuration phase. If there's some way to get away without iterating over the entire thing, I think that would be optimal.

Tapchicoma: Changed to `List<String>` in extension, so applier creates a `FileTree` from provided paths.

Tapchicoma:
@JLLeitschuh can you recheck this PR again?

Tapchicoma:
Yes, I will make a new release and then apply it on plugin project.

PR #218, opened 2019-03-14
Fix task failed to apply

### LINKED ISSUE #217: 7.1.0 - > 7.2.0
Bumped version from 7.1.0 to 7.2.0 and getting:

No value has been specified for property 'android$ktlint_gradle'.

Am using kotlin jvm.  

Build is here: https://github.com/nwillc/ksvg/blob/master/build.gradle.kts

Tapchicoma:
Fixes #217 

Includes setup for `7.2.1` release.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -285,7 +285,6 @@ open class KtlintPlugin : Plugin<Project> {
             description = "Runs a check against all .kt files to ensure that they are formatted according to ktlint."
             configurePluginTask(target, extension, ktLintConfig) {
                 setSource(kotlinSourceDirectories)
-                android.set(extension.android)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -382,6 +382,7 @@ open class KtlintPlugin : Plugin<Project> {
             description = "Format all .kts files in project directory " +
                 "to ensure that they are formatted according to ktlint."
             configurePluginTask(target, extension, ktLintConfig) {
+                android.set(false)
```
JLLeitschuh: Why would someone want to not use the android format for their scripts?
Shouldn't we just honor the flag universally?

Tapchicoma: Good question :thinking: My opinion was somehow biased that android style guide should be only applied for android project sources, but not for kotlin script files.

I would remove that.

JLLeitschuh: I think I agree that removing it and going with the universal configuration is better. Alternatively, offer a separate configuration flag.

Tapchicoma: Removed this.

Tapchicoma:
Just need to update Gradle in plugin and samples to the latest version, test is not needed.

Issue #101, opened 2018-06-08
Add task to install git pre-commit-hook


Tapchicoma:
Ktlint [supports](https://github.com/shyiko/ktlint/blob/master/CHANGELOG.md#0100---2017-10-10) installing git pre-commit hook, that will check style violations on git commit. Would be nice to add this as a Gradle task to simplify setup for the project.

Name should be: `ktlintAddGitPreCommitHook`

JLLeitschuh:
Sounds like a sold plan. How are inputs handled? Eg. custom rulesets?

Also, once installed, this would be somewhat out-of-band from our plugin. Are we okay with that?

There are other solutions that exist for installing commit hooks with Gradle.
https://plugins.gradle.org/search?term=git-hook

Tapchicoma:
That are good questions :+1: 

Regarding them I am thinking of adding two possible git hooks:
- runs Gradle task `ktlintCheck` before commit
- runs Gradle task `ktlintFormat` before commit

This way git hook will be not detached from plugin and plugin configuration.

ubuntudroid:
@Tapchicoma any news here? Would be great to have that to make set up as easy as it is for ktlint.

Tapchicoma:
@ubuntudroid It is one of possible issues that I may work for next release.

JLLeitschuh:
@Tapchicoma This would make sense to add as a standalone plugin, similar to `ktlint-idea`.

Tapchicoma:
@JLLeitschuh not sure if this will be possible if this git hook will call `ktlintCheck` or `ktlintFormat` tasks from main plugin.

JLLeitschuh:
If you call `ktlint --install-git-pre-commit-hook` what does it do?
I thought that was all this proposed addition would do, (although I'm not 100% certain that it would work in a Gradle environment).

If you're planning on having this Gradle task get executed, I think that might be outside the scope of this plugin.

I think that this goal better achieved with a plugin specific for installing pre-commit hooks.

https://plugins.gradle.org/search?term=git-hook

If you want to add a sample that demonstrates using one of these plugins, that would be an acceptable addition to this repo.

ubuntudroid:
`ktlint --install-git-pre-coomit-hook` simply creates a file called pre-commit in .git/hooks which contains the code to run the checks. 

This hook uses ktlint directly so we would need our own variant of the command as one of the benefits from this gradle plugin is that users don't have to have raw ktlint installed on their machine.

[...]

PR #221, opened 2019-04-07
Add pre commit Git hook

Tapchicoma:
Add two tasks that install Git `pre-commit` hook:
- `addKtlintCheckGitPreCommitHook`
- `addKtlintFormatGitPreCommitHook`

Hook itself collects all staged (going to be commited) `*.kt` or `*.kts` files and run over this files either ktlint check or format tasks, passing list of files using `-PinternalKtlintGitFilter` property.

This PR partially addresses #101. Currently there are no support for composite builds and `buildSrc` sub-projects. This support should be added in followup PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-  - ?
+  - Git pre-commit hook (#101):
+
+    Current implementation does not support `buildSrc` or composite builds.
 ### Changed
   - ?
 ### Removed
@@ -32,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [7.1.0] - 2019-02-05
 ### Added
+
   - Warning about using vulnerable ktlint version
 ### Changed
   - Default ktlint version is set to `0.30.0`
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
@@ -1,0 +1,155 @@
+package org.jlleitschuh.gradle.ktlint
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.Project
+import org.gradle.api.file.ProjectLayout
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import org.intellij.lang.annotations.Language
+import java.io.File
+import javax.inject.Inject
+
+internal const val FILTER_INCLUDE_PROPERTY_NAME = "internalKtlintGitFilter"
+
+@Language("Bash")
+internal val shShebang = """
+#!/bin/sh
+set -e
+
+""".trimIndent()
+
+internal const val startHookSection = "######## KTLINT-GRADLE HOOK START ########\n"
+internal const val endHookSection = "######## KTLINT-GRADLE HOOK END ########\n"
+
+@Language("Bash")
+internal fun generateGitHook(taskName: String) = """
+
+CHANGED_FILES="${'$'}(git --no-pager diff --name-status --no-color --cached | awk '$1 != "D" && $2 ~ /\.kts|\.kt/ { print $2}')"
+
+if [[ -z "${'$'}CHANGED_FILES" ]]; then
+    echo "No Kotlin staged files."
+    exit 0
+fi;
+
+echo "Running ktlint over these files:"
+echo "${'$'}CHANGED_FILES"
+
+./gradlew --quiet $taskName -P$FILTER_INCLUDE_PROPERTY_NAME="${'$'}CHANGED_FILES"
+
+echo "Completed ktlint run."
+
+while read -r file; do
+    if [[ -f ${'$'}file ]]; then
+        git add ${'$'}file
+    fi
+done <<< "${'$'}CHANGED_FILES"
+
+echo "Completed ktlint hook."
+
+""".trimIndent()
+
+internal fun KtlintPlugin.PluginHolder.addGitHookTasks() {
+    if (target.rootProject == target &&
+        target.checkGitIsPresent()) {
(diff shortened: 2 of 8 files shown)


Tapchicoma:
@JLLeitschuh I am going to add tests for it, but implementation wise - it is ready.

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-  - ?
+  - Git pre-commit hook (#101)
+  Current implementation does not support `buildSrc`/composite builds.
```
JLLeitschuh: Is this indented correctly?

Also:
> Current implementation does not support `buildSrc`/composite builds.

I'd change this to:
> Current implementation does not support `buildSrc` or composite builds.

JLLeitschuh:
Any concern about having too many args (or an argument that is too long) passed to a CLI? If so, how can we remedy that?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -0,0 +1,152 @@
...
+
+if [[ -z "${'$'}CHANGED_FILES" ]]; then
+    echo "No Kotlin staged files."
+    exit 0
+fi;
+
+echo "Running ktlint over these files:"
+echo "${'$'}CHANGED_FILES"
+
+./gradlew --quiet $taskName -P$FILTER_INCLUDE_PROPERTY_NAME="${'$'}CHANGED_FILES"
```
JLLeitschuh: Are there any weird file names that would cause this to blow up? Eg. File names with spaces in them?

Tapchicoma: Having spaces in file names is ok, as this property is split using `'\n'` (end of line) character.

And if there are such file names - I would not do here premature optimization, but would wait for issues. Githook install tasks have hook-update mechanism on new run, so it will be easy to update git hook on new version.

Tapchicoma:
> Any concern about having too many args (or an argument that is too long) passed to a CLI? If so, how can we remedy that?

Possible, but this cannot be avoided in this case. I prefer to not add some java git client as a dependency here.

JLLeitschuh:
What happens when you run the install git hook task twice? Will the hook code be added twice?

Also, does this work on Windows?

Tapchicoma:
> What happens when you run the install git hook task twice? Will the hook code be added twice?

No, it will overwrite everything between `######## KTLINT-GRADLE HOOK START ########\n` and `######## KTLINT-GRADLE HOOK END ########\n` lines. If there is some other hooks already present - task will append ktlint-gradle hook in the end of file.

> Also, does this work on Windows?

I am not a Windows user. According to some answers:
"By default, Git for Windows executes hook scripts using its own Windows port of the bash shell." Though, I think, because of `awk` usage - it will not work.
Would be happy to accept PR from a Windows user.

Tapchicoma:
From my point of view - this PR is ready to be merged.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/GitHookTasksTest.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import java.io.File
+
+class GitHookTasksTest : AbstractPluginTest() {
+    @BeforeEach
+    internal fun setUp() {
+        //language=Groovy
```
JLLeitschuh: Nice!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.gradle.api.tasks.TaskAction
+import org.intellij.lang.annotations.Language
+import java.io.File
+import javax.inject.Inject
+
+internal const val FILTER_INCLUDE_PROPERTY_NAME = "internalKtlintGitFilter"
+
+@Language("Bash")
+internal val shShebang = """
+#!/bin/sh
```
Globegitter: @Tapchicoma I had to change this to `#!/bin/bash` on my machine, which is Elementary OS 5 / Ubuntu 18.04 - do you think it makes sense to change this to bash?

Tapchicoma: Why do you need this change? :thinking:

JLLeitschuh: I'm guessing it's because Elementary  OS doesn't ship with `sh`, only `bash`.

Tapchicoma: Seems so: https://askubuntu.com/questions/537669/script-doesnt-work-when-it-is-run-using-bin-sh

I will check what bashisms this script is using.

Tapchicoma:
```
$ checkbashisms .git/hooks/pre-commit
possible bashism in .git/hooks/pre-commit line 7 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ -z "$CHANGED_FILES" ]]; then
possible bashism in .git/hooks/pre-commit line 20 (alternative test command ([[ foo ]] should be [ foo ])):
    if [[ -f $file ]]; then
possible bashism in .git/hooks/pre-commit line 23 (<<< here string):
done <<< "$CHANGED_FILES"
```

Globegitter: I do actually have a `/bin/sh` on my disk but I guess it is some less advanced shell. And yes it was failing on line 23 `done <<< "$CHANGED_FILES"`

Globegitter: Yeah my `sh` corresponds to `dash`.

Tapchicoma: I will try to fix this in a separate PR. Thank you for letting us know it :)

PR #228, opened 2019-04-23
Add support for pinterest package and group changes

### LINKED ISSUE #227: Update `ktlint` to "0.32.0"
Due to changes in group id and package makes sense to update the dependencies.

https://github.com/pinterest/ktlint/blob/master/CHANGELOG.md#0320---2019-04-22

tasomaniac:
### Problem 

ktlint just released version 0.32.0 to fix a problem with Kotlin 1.3.30. But at the same time, they also changed the maven group id and package names. That is why just updating ktlint version to `0.32.0` breaks ktlint plugin.

### Solution

I've replaces groups and packages to support the new version. These changes here should be backward compatible.

### Considerations

Since this is only required when using version 0.32.0, I haven't changed the sample repo.
Added test that checks that the correct dependency is added.

Fixes #227

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintApplyToIdeaTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintApplyToIdeaTask.kt
@@ -1,6 +1,5 @@
 package org.jlleitschuh.gradle.ktlint
 
-import net.swiftzer.semver.SemVer
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.model.ObjectFactory
@@ -31,7 +30,7 @@ open class KtlintApplyToIdeaTask @Inject constructor(
     fun generate() {
         project.javaexec {
             it.classpath = classpath
-            it.main = resolveMainClassName()
+            it.main = resolveMainClassName(ktlintVersion.get())
             if (globally.get()) {
                 it.args("--apply-to-idea")
             } else {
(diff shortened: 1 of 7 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -143,10 +151,11 @@ internal fun String.androidVariantMetaFormatTaskName(
...
  * Get specific android variants from [BaseExtension].
  */
-internal val BaseExtension.variants: DomainObjectSet<out BaseVariant>? get() = when (this) {
-    is AppExtension -> applicationVariants
-    is LibraryExtension -> libraryVariants
-    is FeatureExtension -> featureVariants
-    is TestExtension -> applicationVariants
-    else -> null // Instant app extension doesn't provide variants access
-}
+internal val BaseExtension.variants: DomainObjectSet<out BaseVariant>?
```
tasomaniac: I have the official Kotlin code style applied and it forces me to have the `get()` in a new line.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintApplyToIdeaTask.kt
```diff
@@ -43,4 +44,9 @@ open class KtlintApplyToIdeaTask @Inject constructor(
             }
         }
     }
+
+    private fun resolveMainClassName() = when {
```
Tapchicoma: Would make sense to move this method to `PluginUtil` file.

Tapchicoma: I see you've added it there - then just make it `internal` and use it.

Tapchicoma:
@tasomaniac can you also update `CHANGELOG.md` with:

```markdown
## [Unreleased]
### Changed
  - Default ktlint version to `0.32.0`. 

  Add support different group/package name (#228)
```

PR #235, opened 2019-04-25
New dev cycle

Tapchicoma:
- Updated versions.
- Update Gradle to `5.1.1` version in samples (blocker: https://github.com/JetBrains/kotlin-native/issues/2916) and to `5.4` in plugin project.
- Simplify plugin setup in samples.


build.gradle.kts
```diff
@@ -4,16 +4,13 @@ buildscript {
...
-        classpath("org.jlleitschuh.gradle:ktlint-gradle:+")
-    }
 }
 
 plugins {
     kotlin("jvm") version SamplesVersions.kotlin apply false
     id("com.android.application") version SamplesVersions.androidPlugin apply false
-    id("konan") version SamplesVersions.kotlin apply false
+    id("org.jetbrains.kotlin.konan") version SamplesVersions.kotlin apply false
+    id("org.jlleitschuh.gradle.ktlint")
```
JLLeitschuh: ※\\(^o^)/※

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-bin.zip
```
JLLeitschuh: Should be `ALL` distribution right?

JLLeitschuh:
Does this break backward compatibility with the android/kotlin plugins?

gradle/wrapper/gradle-wrapper.properties
Tapchicoma: This wrapper is used for samples project. Plugin project (that is included into samples project as composite build) uses `ALL` distribution.

Tapchicoma:
> Does this break backward compatibility with the android/kotlin plugins?

Should not.

PR #240, opened 2019-05-04
Change tasks inheritance.

### LINKED ISSUE #225: Consider breaking inheritance hierarchy from `KtlintCheckTask` and `KtlintFormatTask`
https://github.com/JLLeitschuh/ktlint-gradle/blob/90b0c63d3e6dac9a46b85a777a287553917fab8d/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintFormatTask.kt#L14-L16

https://github.com/JLLeitschuh/ktlint-gradle/blob/90b0c63d3e6dac9a46b85a777a287553917fab8d/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt#L33-L35

When doing something like

```kotlin
withType<KtlintCheckTask> {
  // use task
}
```

It was unexpected for me to also get all of the `KtlintFormatTask` tasks and configure them too.

Tapchicoma:
Now `KtlintCheck` and `KtlintFormat` tasks extends base `BaseKtlintCheckTask`.

Closes #225


Tapchicoma:
I would say that it is not api breaking change as tasks package/names stays the same - only difference now that to configure all the tasks - consumer should use `BaseKtlintCheckTask` instead of `KtlintFormatTask`.

JLLeitschuh:
Should we call it `BaseKtlintCheckTask`? Is it really a check? Should we call it `BaseKtlintTask`?

It's technically a breaking change because you're changing the inheritance hierarchy for one of the tasks so `tasks.withType` won't configure all of our tasks. Am I understanding the change correctly?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintFormatTask.kt
```diff
@@ -15,7 +15,7 @@ import javax.inject.Inject
 open class KtlintFormatTask @Inject constructor(
     objectFactory: ObjectFactory,
     projectLayout: ProjectLayout
-) : KtlintCheckTask(objectFactory, projectLayout) {
+) : BaseKtlintCheckTask(objectFactory, projectLayout) {
```
JLLeitschuh: I think this is an API breaking change because `tasks.withType<KtlintCheckTask>` will now behave differently.

Tapchicoma:
> Should we call it BaseKtlintTask?

But there are also tasks that do not perform checks :slightly_smiling_face: Naming is hard :sweat: 

I am fine with going `8.0.0` release, no strong opinion here.

PR #247, opened 2019-06-09
Add support for additional editorconfig file

### LINKED ISSUE #210: Add support for editorconfig path

Tapchicoma:
Closes #210

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
@@ -37,12 +37,23 @@ internal inline fun <reified T : Task> Project.registerTask(
 
 internal const val EDITOR_CONFIG_FILE_NAME = ".editorconfig"
 
-internal fun getEditorConfigFiles(currentProject: Project): FileCollection {
-    return searchEditorConfigFiles(
+internal fun getEditorConfigFiles(
+    currentProject: Project,
+    additionalEditorconfigFile: RegularFileProperty
+): FileCollection {
+    var editorConfigFileCollection = searchEditorConfigFiles(
         currentProject,
         currentProject.projectDir.toPath(),
         currentProject.files()
     )
+
+    if (additionalEditorconfigFile.isPresent) {
+        editorConfigFileCollection = editorConfigFileCollection.plus(
+            currentProject.files(additionalEditorconfigFile.asFile.get().toPath())
+        )
+    }
+
+    return editorConfigFileCollection
 }
 
 private tailrec fun searchEditorConfigFiles(
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
@@ -1,0 +1,111 @@
+package org.jlleitschuh.gradle.ktlint
+
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+/**
+ * Runs [EditorConfigTests] with the current version of Gradle.
+ */
+class GradleCurrentEditorConfigTest : EditorConfigTests()
+
+/**
+ * Runs [EditorConfigTests] with lowest supported Gradle version.
+ */
+@Suppress("ClassName")
+class Gradle4_10EditorConfigTest : EditorConfigTests() {
+
+    override fun gradleRunnerFor(vararg arguments: String): GradleRunner =
+        super.gradleRunnerFor(*arguments).withGradleVersion("4.10")
+}
+
+/**
+ * Contains all tests related to `.editorconfig` files support.
+ */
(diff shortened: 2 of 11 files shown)


JLLeitschuh:
Can you specify only one or does it support multiple files?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -37,12 +37,23 @@ internal inline fun <reified T : Task> Project.registerTask(
 
 internal const val EDITOR_CONFIG_FILE_NAME = ".editorconfig"
 
-internal fun getEditorConfigFiles(currentProject: Project): FileCollection {
-    return searchEditorConfigFiles(
+internal fun getEditorConfigFiles(
+    currentProject: Project,
+    additionalEditorconfigFile: RegularFileProperty
+): FileCollection {
+    var editorConfigFileCollection = searchEditorConfigFiles(
```
JLLeitschuh: Should we even search if the user has specified their own? If there is both one in the repository and one passed via the command line, which one does ktlint use?

Tapchicoma: According to ktlint code - it is still using working directory `.editorconfig` file even if `--editorconfig` option was used:
- loading editorconfig passed via `--editorconfig` option:
https://github.com/pinterest/ktlint/blob/af7202536fd76cce9b3bb6ff0f81855de5060f3e/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt#L379
- loading current working directory editorconfigs:
https://github.com/pinterest/ktlint/blob/af7202536fd76cce9b3bb6ff0f81855de5060f3e/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt#L387

`editorConfigFileCollection` is just used as a task input, but never passed to the ktlint itself, so merging here all editorconfig files is right behaviour.

Though, actually, I've noticed that in this PR additional `.editorconfig` file was never passed to ktlint itself :+1:

Tapchicoma:
> Can you specify only one or does it support multiple files?

Actually type of the extension field is `RegularFileProperty`, so only one file could be set.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
Tapchicoma: Fixed in https://github.com/JLLeitschuh/ktlint-gradle/pull/247/commits/eab563de0cfa9d561b32e00d77a7f427d463f903

JLLeitschuh: If it's an override, then it should be called an override, not `additional`. `additional` sounds like they are being merged or something.

Tapchicoma: Sorry, maybe I was not clear - ktlint adds additional `.editorconfig` file to the `.editorconfig` files that are detected in the working directory.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -37,12 +37,23 @@ internal inline fun <reified T : Task> Project.registerTask(
...
         currentProject,
         currentProject.projectDir.toPath(),
         currentProject.files()
     )
+
+    if (additionalEditorconfigFile.isPresent) {
+        editorConfigFileCollection = editorConfigFileCollection.plus(
+            currentProject.files(additionalEditorconfigFile.asFile.get().toPath())
+        )
```
JLLeitschuh: Ohhhh. I didn't spot this was an `add`. That's why I was so confused.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
JLLeitschuh: I think I understand now. I didn't see that it's an `add` below. 🤦‍♂

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/TestsCommon.kt
```diff
@@ -0,0 +1,31 @@
...
+    //language=Groovy
+    buildFile().writeText(
+        """
+            $pluginsBlockWithMainPluginAndKotlinJvm
+
+            repositories {
+                gradlePluginPortal()
+            }
+
+            import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
```
JLLeitschuh: 🤔 Didn't know you could have an import statement in the middle of a file in groovy scripts. Can you do the same thing in Kotlin?! Huh.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
```diff
@@ -0,0 +1,167 @@
+package org.jlleitschuh.gradle.ktlint
```
JLLeitschuh: Good idea extracting this to a separate test.

PR #255, opened 2019-06-23
Fix bashisms in git hook.

### LINKED ISSUE #251: Git hook has bashisms
We need to fix this bashisms to actually work under `/bin/sh` shebang:
```
$ checkbashisms .git/hooks/pre-commit
possible bashism in .git/hooks/pre-commit line 7 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ -z "$CHANGED_FILES" ]]; then
possible bashism in .git/hooks/pre-commit line 20 (alternative test command ([[ foo ]] should be [ foo ])):
    if [[ -f $file ]]; then
possible bashism in .git/hooks/pre-commit line 23 (<<< here string):
done <<< "$CHANGED_FILES"
```

_Originally posted by @Tapchicoma in https://github.com/JLLeitschuh/ktlint-gradle/pull/221_

Tapchicoma:
This will allow to use in pure sh shells implementations.

Running `checkbashisms .git/hooks/pre-commit` produces no warnings.

Fixes #251

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: android
 jdk:
   - oraclejdk8
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq devscripts
 install:
   - yes | sdkmanager --verbose "build-tools;26.0.2"
   - yes | sdkmanager --verbose "platform-tools"
@@ -12,6 +15,7 @@ script:
   - ./plugin/gradlew -p ./plugin build --no-daemon
   - ./plugin/gradlew -p ./plugin ktlintCheck --no-daemon
   - ./gradlew ktlintCheck --no-daemon
+  - ./gradlew addKtlintCheckGitPreCommitHook --no-daemon && checkbashisms .git/hooks/pre-commit
   - if [ -n "$TRAVIS_TAG" ]; then ./plugin/gradlew -p ./plugin publishPlugins --no-daemon; fi
 
 before_cache:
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
Can we run `checkbashisms` as part of our tests?

Tapchicoma:
Theoretically - yes, it is just perl script.

Tapchicoma:
Added such check in https://github.com/JLLeitschuh/ktlint-gradle/pull/255/commits/3cefdd0022bf065d0dfc749082d89c03543f2c33

.travis.yml
```diff
@@ -12,6 +15,7 @@ script:
   - ./plugin/gradlew -p ./plugin build --no-daemon
   - ./plugin/gradlew -p ./plugin ktlintCheck --no-daemon
   - ./gradlew ktlintCheck --no-daemon
+  - ./gradlew addKtlintCheckGitPreCommitHook --no-daemon && checkbashisms .git/hooks/pre-commit
```
JLLeitschuh: Just sanity checking `checkbashisms .git/hooks/pre-commit` will return a non-zero exit code if this fails thus failing Travis CI right?

Tapchicoma: according to man page:
```
EXIT VALUES
       The exit value will be 0 if no possible bashisms or other problems were detected.  Otherwise it will be the sum of the following error values:

       1      A possible bashism was detected.

       2      A file was skipped for some reason, for example, because it was unreadable or not found.  The warning message will give details.

       4      No bashisms were detected in a bash script.
```

PR #256, opened 2019-06-26
Add support for new "org.jetbrains.kotlin.js" plugin.

### LINKED ISSUE #252: Add support for new JS Gradle plugin
With Kotlin release `1.3.40` new JS Gradle plugin was [introduced](https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/): https://plugins.gradle.org/plugin/org.jetbrains.kotlin.js

Our plugin should add support for this plugin as well.

Tapchicoma:
Closes #252

### DIFF
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
@@ -1,0 +1,73 @@
+package org.jlleitschuh.gradle.ktlint
+
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+/**
+ * Runs [KotlinJsPluginTests] with the current version of Gradle.
+ */
+class GradleCurrentKotlinJsPluginTests : KotlinJsPluginTests()
+
+/**
+ * Runs [KotlinJsPluginTests] with lowest supported Gradle version.
+ */
+@Suppress("ClassName")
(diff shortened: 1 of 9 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
```diff
@@ -0,0 +1,73 @@
+package org.jlleitschuh.gradle.ktlint
```
JLLeitschuh: Thanks for staying on top of these compatibility tests and keeping them refactored as they need to be changed.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
```
JLLeitschuh: This is a nice API that `assertThat` proivides. I may have to give it a try.

samples/kotlin-kotlin2js/src/main/kotlin/org/jlleitschuh/gradle/ktlint/sample/js/main.kt
```diff
@@ -0,0 +1,14 @@
+package org.jlleitschuh.gradle.ktlint.sample.js
+
+import kotlin.browser.document
+
+val secondDiv: dynamic = document.createElement("div").apply {
```
JLLeitschuh: Forgive my lack of experience. What is the `dynamic` type in Kotlin? This is the first I've ever seen it.

Tapchicoma: It is only js specific: https://kotlinlang.org/docs/reference/dynamic-type.html

PR #258, opened 2019-07-01
Add support for custom ktlint reporters

### LINKED ISSUE #125: Add support for custom reporters.
See https://github.com/shyiko/ktlint#creating-a-reporter

Tapchicoma:
Closes #125


Tapchicoma:
@JLLeitschuh could you check general approach in this PR?

README.md
```diff
@@ -168,6 +168,9 @@ ktlint {
         exclude("**/generated/**")
         include("**/kotlin/**")
     }
+    customReporters {
```
Tapchicoma: Possibly this syntax may be used for built-in reporters as well:
```gradle
reporters {
   repoter "plain"
   customReporter "csv", "csv", project(":csv-reporter")
}
```

README.md
```diff
@@ -168,6 +168,9 @@ ktlint {
...
+        reporter "csv", "csv", project(":csv-reporter")
```
JLLeitschuh: Oh, wow, that's impressive for DSL/API support.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
```diff
@@ -83,6 +89,33 @@ abstract class BaseKtlintCheckTask(
             }
             .filter { it.enabled.get() }
 
+    @get:Internal
+    internal val customReports
```
JLLeitschuh: For my sanity, can you be explicit about the type here?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
```diff
@@ -73,7 +78,8 @@ abstract class BaseKtlintCheckTask(
     internal val enabledReports
```
JLLeitschuh: Same question here, can you help my sanity and add an explicit type here?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
```diff
@@ -382,6 +384,9 @@ open class KtlintPlugin : Plugin<Project> {
 
         val ktlintConfiguration by lazy(LazyThreadSafetyMode.NONE) { createKtlintConfiguration(target, extension) }
         val ktlintRulesetConfiguration = createKtlintRulesetConfiguration(target)
+        val ktlintReporterConfiguration by lazy(LazyThreadSafetyMode.NONE) {
```
JLLeitschuh: Isn't that the default thread safety mode?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/CustomReporter.kt
```diff
@@ -0,0 +1,17 @@
...
+    val reporterId: String,
+    val reporterFileExtension: String,
+    val reporterDependencyNotation: Any
+) {
+    internal lateinit var dependency: Dependency
+
+    internal fun resolveDependency(project: Project): Dependency {
+        dependency = project.dependencies.create(reporterDependencyNotation)
+        return dependency
+    }
```
JLLeitschuh: Any better way to do this without `lateinit`?
Like what happens if one `CustomReporter` is shared between two projects? I don't know if that's even possible, but it if is, you could end up with a `Dependency` resolved against the wrong project perhaps?

samples/kotlin-reporter-creating/src/main/resources/META-INF/services/com.pinterest.ktlint.core.ReporterProvider
```diff
@@ -0,0 +1 @@
+org.jlleitschuh.gradle.ktlint.sample.CsvReporterProvider
```
JLLeitschuh: Newline at EOF?

JLLeitschuh:
Can you put a README in `samples/kotlin-reporter-creating` describing the structure? Would it help?

Do we need to explain to developers that if they side effect change other files outside of the input and output we won't see their changes and it won't get captured in Gradle's "out of date" checking.

Also, just checking, how do custom reporters work with the Gradle Build Cache?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
Tapchicoma: Nope, default is `LazyThreadSafetyMode.SYNCHRONIZED`

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/CustomReporter.kt
Tapchicoma: Actually it is not needed, removed.

README.md
Tapchicoma: @JLLeitschuh any opinion on this?

Tapchicoma:
> Can you put a README in samples/kotlin-reporter-creating describing the structure? Would it help?

Maybe I misunderstand you, but why we need to add README how to create ktlint reporter? It should be in a ktlint project.

> Do we need to explain to developers that if they side effect change other files outside of the input and output we won't see their changes and it won't get captured in Gradle's "out of date" checking.

Can you explain here?

> Also, just checking, how do custom reporters work with the Gradle Build Cache?

Should work, as all reporters produces task `@OutputFiles` property. I will add additional tests for it.

README.md
JLLeitschuh: Can you explain them both better please so I know what I'm picking?

Tapchicoma: Currently to set built-in reporter plugin user calls:
```kotlin
    reporters.set(setOf(ReporterType.PLAIN, ReporterType.CHECKSTYLE))
```

My idea is to move this to `reportersDsl { .. }` block, so setting reporters would be unified (see previous comment how it may look).

JLLeitschuh: Does the DSL also allow the developer to describe what the reporter's output type is?

JLLeitschuh:
> > Do we need to explain to developers that if they side effect change other files outside of the input and output we won't see their changes and it won't get captured in Gradle's "out of date" checking.

> Can you explain here?

For example:

If you side effect other files in your report, you will need to manually add those side affected files as outputs of the Gradle task. Failure to do so will mean that Gradle will be unable to fully detect all file changed by the task and thus may incorrectly mark your tasks as "up-to-date" when they actually aren't.

README.md
Tapchicoma: For built-in reporters developer can't define output type, same as it is already via `reporters.set(..)` call.

Tapchicoma:
> If you side effect other files in your report, you will need to manually add those side affected files as outputs of the Gradle task. Failure to do so will mean that Gradle will be unable to fully detect all file changed by the task and thus may incorrectly mark your tasks as "up-to-date" when they actually aren't.

Sorry, still don't get it fully - how can consume "side-effect" output file?

All 3rd party output reports are added to task outputs, so if they will be removed or modified - task will re-run again.

JLLeitschuh:
Example:
```
class CsvReporter(
    private val out: PrintStream
) : Reporter {
    override fun onLintError(file: String, err: LintError, corrected: Boolean) {
        val line = "$file;${err.line};${err.col};${err.ruleId};${err.detail};$corrected"
        out.println(line)
        File("some_other_file.txt").write(line) // Side effect: `some_other_file.txt` won't be captued as an "output"
    }
}
```

Tapchicoma:
Now I got it. Yeah, such "side effect" should be mentioned in README. Not sure if it should be in this project or ktlint? :thinking:

JLLeitschuh:
> Yeah, such "side effect" should be mentioned in README. Not sure if it should be in this project or ktlint?

Not sure. Both?

README.md
JLLeitschuh: Can you give an example of the DSL usage with only a single custom reporter added?

Tapchicoma: Only one custom reporter added:
```gradle
reporters {
   customReporter "csv", "csv", project(":csv-reporter")
}
```

Only built-in reporter added:
```gradle
reporters {
   repoter "plain"
}
```

JLLeitschuh: In the documentation can we make it:

```kotlin
reporters {
   customReporter "csv-reporter", ".csv", project(":csv-reporter")
}
```

That way it's really clear what's going on for each parameter?

JLLeitschuh: Also, you probably want to accept a `Configuration` as the last argument instead of a project, right?

Tapchicoma: First parameter of custom reporter is the id of the reporter, so it should stay "csv": https://github.com/Tapchicoma/ktlint-gradle/blob/ce38cea8af5654c60b924329f4d353d1aac7c042/samples/kotlin-reporter-creating/src/main/kotlin/org/jlleitschuh/gradle/ktlint/sample/CsvReporterProvider.kt#L8

Second parameter is generated reporter file extension, it should always contain '.' symbol, so, imho, better to omit it.

last parameter is "dependency notation" from Gradle [DependencyHandler](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/DependencyHandler.html), that is added by plugin to internal "ktlintReporter" configuration.

JLLeitschuh: I guess the point that I'm trying to make is that from the perspective of a lay-person, the API is somewhat confusing and unclear what each parameter is.

May I suggest an alternative API?

```kotlin
reporters {
    customReporters {
        "csv" {
            extension = "csv" // This could also be implied from the name
            from(project(":csv-reporter"))
        }
    }
}
```

In order to accomplish this API, you could make the `customReporters` a `NamedDomainObjectContainer`.

Simply making `customReporters` a `NamedDomainObjectContainer<CustomReporter>` where `CustomReporter` is something like:

```kotlin
data class CustomReporter(
    val name: String,
    var extension: String = name,
    var from: [Something]? = null
)
```

I know this is a lot to change, but I believe the API is clearer, more flexible, and is more "Gradle like".

Tapchicoma: Hm, looks interesting. Let me play with it. I suppose built-in reporters would have proposed above syntax:
```kotlin
reporters {
    reporter "checkstyle"
}
```
?

JLLeitschuh: You could keep the built in ones as simple as you want.

AdamMc331:
Is this still in development? I was looking into the idea of a custom reporter to tweak the colored output for KtLint, being discussed here: https://github.com/pinterest/ktlint/issues/582 

Am I unable to implement a custom reporter without this PR being merged?

Tapchicoma:
Yes, I am planning to finish it before next release (pretty soon). Until then, unfortunately, you can't use custom reporter with this plugin.

AdamMc331:
Thanks for the update! Ping me if I can provide assistance in any way. :)

README.md
Tapchicoma: @JLLeitschuh please check this commit https://github.com/JLLeitschuh/ktlint-gradle/pull/258/commits/33d510c5323deaeef1d8a8afa83ea07ef9ad0dc8 if this new DSL looks good for you.

Tapchicoma:
@JLLeitschuh this PR is ready to be reviewed again

PR #263, opened 2019-07-17
Add support for "com.android.dynamic-feature" plugin.

### LINKED ISSUE #260: Plugin doesn't work for dynamic-feature modules
In root build.gradle I have a configuration :

```gradle
dependencies {
    classpath "org.jlleitschuh.gradle:ktlint-gradle:8.1.0"
}
subprojects {

    // Optionally configure plugin
    ktlint {
        debug = true
        verbose = true
        outputToConsole = true
        android = true
    }
}
```

But kotlinCheck task works only for the main app module.
Seems the plugin skips sources for all non main modules.

Any suggestions?

Tapchicoma:
Closes #260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 ### Added
   - Support for new JS plugin: "org.jetbrains.kotlin.js" (#252)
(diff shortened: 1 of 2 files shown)


Tapchicoma:
@JLLeitschuh merging as here just a small improvement.

JLLeitschuh:
Pretty awesome how, because this plugin is so well laid out, you can just drop in new plugins like this that easily.

That's pretty awesome yo!

PR #280, opened 2019-09-10
Add support to disable particular rules via extension property

### LINKED ISSUE #267: Support `--disabled_rules`
Please support `--disabled_rules`.

I need it to disable `no-wildcard-imports`, because IntelliJ keeps adding in the dreaded `java.util.*` due to its misguided `Packages to Use Import with '*'` feature.

Tapchicoma:
Closes #267.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
```diff
@@ -147,6 +150,12 @@ abstract class BaseKtlintCheckTask(
             enabledReports
                 .map { it.asArgument() }
                 .forEach { argsWriter.println(it) }
+            disabledRules
+                .get()
+                .fold("") { acc, rule -> "$acc,$rule" }
```
JLLeitschuh: I think you want `joinToString` or something here. You can pass a seperator.

Tapchicoma: You are right! Totally forgot about it :man_facepalming:

PR #281, opened 2019-09-10
Enable using the latest master snapshot version of ktlint

hosamaly:
I was trying to follow the instructions to [use the latest `master` snapshot of ktlint](https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot) today, but I saw the error about versions less than 0.22 not being supported.

This PR enables the use of master snapshots and treats it as the latest version, even though it's original version is 0.0.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -23,8 +23,16 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.gradle.util.GradleVersion
 import java.nio.file.Path
 
+// The latest commit on ktlint's master branch is automatically uploaded to Sonatype's snapshots repository
+// https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot
+private const val KtlintMasterSnapshotVersion = "0.0.0-SNAPSHOT"
```
JLLeitschuh:
```suggestion
private const val ktlintMasterSnapshotVersion = "0.0.0-SNAPSHOT"
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -23,8 +23,16 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin
...
+
+internal fun ktlintVersionLessThan(expectedVersion: SemVer, ktlintVersion: String) =
+    ktlintVersion != KtlintMasterSnapshotVersion &&
```
JLLeitschuh:
```suggestion
    ktlintVersion != ktlintMasterSnapshotVersion &&
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -23,8 +23,16 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.gradle.util.GradleVersion
 import java.nio.file.Path
 
+// The latest commit on ktlint's master branch is automatically uploaded to Sonatype's snapshots repository
+// https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot
```
JLLeitschuh: Can you turn these two lines into a kotlin doc comment instead?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
hosamaly: Done in 24d2032

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
hosamaly: Done in 24d2032

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
hosamaly: Done in 24d2032

hosamaly:
For some reason, commit 24d203294d3ab9284dc5369d5f8009b7d9d8797c caused the build to fail. I'm fairly new to Kotlin, so I can't see what in the diff could cause that failure. I couldn't reproduce the issue on my machine.

Could you please help me understand what's wrong with that commit?

JLLeitschuh:
I just restarted the build, we'll see what's up.

Tapchicoma:
I would say, that this is more ktlint problem itself. As far as I know it was fixed in Gradle build, but still the issue in the maven builds.

jacquerie:
FWIW I was also hit by this. Is there anything I can do to help get this merged?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -23,8 +23,20 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin
 import org.gradle.util.GradleVersion
 import java.nio.file.Path
 
+/**
+ * The version identifier for ktlint's master snapshots
+ *
+ * The latest commit on ktlint's master branch is automatically uploaded to Sonatype's snapshots repository
+ * https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot
+ */
+private const val ktlintMasterSnapshotVersion = "0.0.0-SNAPSHOT"
```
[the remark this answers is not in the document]
hosamaly: Fixed in f45609f

Tapchicoma:
@hosamaly @jacquerie snapshot with proper versioning was published: https://oss.sonatype.org/content/repositories/snapshots/com/pinterest/ktlint/0.35.0-SNAPSHOT/

jacquerie:
> @hosamaly @jacquerie snapshot with proper versioning was published: https://oss.sonatype.org/content/repositories/snapshots/com/pinterest/ktlint/0.35.0-SNAPSHOT/

Cool, that works for us! Thanks a lot, @Tapchicoma !

Tapchicoma:
@hosamaly sorry for not merging this PR, but proper approach would be to fix snapshots versioning in ktlint repository itself, that is done via https://github.com/pinterest/ktlint/pull/588.

Thank you anyway for this PR and raising this problem!

hosamaly:
No worries. Thank you for fixing the root cause.

PR #289, opened 2019-09-30
Remove unneeded FAQ instructions

tasomaniac:
When #125 was fixed, reports DSL was changed so that these instructions are not necessary anymore.

Note: I haven't verified that this is the case, It looked like so by looking at the changes. Please verify. Thanks.


Tapchicoma:
from my tests this step indeed not needed anymore :+1:

PR #297, opened 2019-10-20
Added output color name property.

AdamMc331:
This has support for the `--color-name` argument added in KtLint 0.35.0: https://github.com/pinterest/ktlint/pull/585

Tested by adding the property to the android-app sample, and forcing a failure.

Before:

Screen Shot 2019-10-20 at 6 48 40 PM

After:

Screen Shot 2019-10-20 at 6 49 16 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 ### Added
-  - ?
+  - Support for outputColorName property (#297)
 ### Changed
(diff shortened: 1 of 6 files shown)


Tapchicoma:
@AdamMc331 could you also update changelog?

AdamMc331:
@Tapchicoma No problem! Do you think this property should also be added to the sample blocks in the README?

Tapchicoma:
> Do you think this property should also be added to the sample blocks in the README?

Yes, would be nice :+1:

AdamMc331:
@Tapchicoma All set! :D Thanks so much for reviewing this.

PR #303, opened 2019-11-10
Disable incremental tasks api for format task.

### LINKED ISSUE #302: Running ktlintFormat twice deletes files
**Steps to reproduce the bug**
1) Use ktlint-gradle version `9.1.0` and ktlint version `0.35.0` with the following configuration:
```
ktlint {
    version = "0.35.0"
    debug = true
    verbose = true
    android = true
    outputToConsole = true
    outputColorName = "RED"
    enableExperimentalRules = false
}
```
2) Run `./gradlew ktlintFormat --continue` once. (Then run `git status`, and observe that the auto-formatted files were modified)
3) Run `./gradlew ktlintFormat --continue` a second time

**Expected Behavior**
Running `./gradlew ktlintFormat --continue` the second time should not making any additional changes. The previously modified files should still show up as modified in `git status`, and no files should be deleted.

**Actual Behavior**
All the files that were previously marked as modified are now `deleted` (observed via `git status`).

I also tried the gradle integration recommended in https://github.com/pinterest/ktlint without plugins, and running the non-plugin `ktlintFormat` task multiple times did not delete any files. This makes me suspect that the bug is in `ktlint-gradle`.

Tapchicoma:
Disabled incremental task api for format task. As format task declares output same as input and incremental api deletes all outputs on inputs changes, this lead to files deletion on the second run.

Fixes #302

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -2,8 +2,12 @@ package org.jlleitschuh.gradle.ktlint
 
 import java.io.PrintWriter
 import javax.inject.Inject
+import org.gradle.api.file.FileType
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.TaskAction
+import org.gradle.work.ChangeType
+import org.gradle.work.InputChanges
 
 @Suppress("UnstableApiUsage")
 @CacheableTask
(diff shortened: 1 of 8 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
```diff
@@ -1,14 +1,38 @@
...
+            .fold(mutableSetOf<File>()) { acc, fileChange ->
+                logger.debug("File changed: $fileChange")
+                with(fileChange) {
+                    if (fileType != FileType.DIRECTORY &&
+                        changeType != ChangeType.REMOVED) {
+                        acc.add(file)
+                    }
+                }
+                acc
+            }
```
JLLeitschuh: You could also do `map` with a `filterNonNull` afterward. This works too though.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -157,14 +157,15 @@ abstract class BaseKtlintPluginTest : AbstractPluginTest() {
...
+            assertThat(task(":ktlintMainSourceSetFormat")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
         }
 
         projectRoot.restoreFailingSources()
 
         build(":ktlintFormat").apply {
-            assertThat(task(":ktlintMainSourceSetFormat")!!.outcome).isEqualTo(TaskOutcome.SUCCESS)
+            assertThat(task(":ktlintMainSourceSetFormat")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
         }
+        assertThat(projectRoot.resolve(FAIL_SOURCE_FILE)).exists()
```
Tapchicoma: @JLLeitschuh the test is here :)

JLLeitschuh: Cool! Awesome!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
Tapchicoma: Force updated this part.

PR #323, opened 2020-01-12
Plugin will search for .git folder starting from the root project folder to install hook.

### LINKED ISSUE #284: Allow to provide non-default path to project .git folder
Followup of #101:
Current implementation only adds a git hook, when `.git/` folder is located in Gradle root project folder. But quite often project has non-gradle top folder. Plugin should add a way to define non-standard `./git` folder location.

Tapchicoma:
Now git tasks are always added to avoid I/O on Gradle configuration phase.

Closes #284

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -1,12 +1,17 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     kotlin("jvm") version PluginVersions.kotlin
     id("com.gradle.plugin-publish") version PluginVersions.gradlePublishPlugin
     `java-gradle-plugin`
     `maven-publish`
     id("org.jlleitschuh.gradle.ktlint") version PluginVersions.ktlintPlugin
+    id("com.github.johnrengelman.shadow") version PluginVersions.shadow
 }
 
-group = "org.jlleitschuh.gradle"
+val pluginGroup = "org.jlleitschuh.gradle"
+group = pluginGroup
 version = "9.2.0-SNAPSHOT"
 
 repositories {
@@ -18,12 +23,20 @@ tasks.withType<PluginUnderTestMetadata>().configureEach {
     pluginClasspath.from(configurations.compileOnly)
 }
 
+/**
+ * Special configuration to be included in resulting shadowed jar, but not added to the generated pom and gradle
+ * metadata files.
+ */
+val shadowImplementation by configurations.creating
+configurations["compileOnly"].extendsFrom(shadowImplementation)
+configurations["testImplementation"].extendsFrom(shadowImplementation)
+
 dependencies {
     compileOnly(gradleApi())
     compileOnly(kotlin("gradle-plugin", PluginVersions.kotlin))
     compileOnly("com.android.tools.build:gradle:${PluginVersions.androidPlugin}")
-    implementation("net.swiftzer.semver:semver:${PluginVersions.semver}")
-    implementation("org.eclipse.jgit:org.eclipse.jgit:${PluginVersions.jgit}")
+    shadowImplementation("net.swiftzer.semver:semver:${PluginVersions.semver}")
+    shadowImplementation("org.eclipse.jgit:org.eclipse.jgit:${PluginVersions.jgit}")
 
     /*
      * Do not depend upon the gradle script kotlin plugin API. IE: gradleScriptKotlinApi()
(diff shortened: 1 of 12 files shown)


plugin/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ dependencies {
     compileOnly(kotlin("gradle-plugin", PluginVersions.kotlin))
     compileOnly("com.android.tools.build:gradle:${PluginVersions.androidPlugin}")
     implementation("net.swiftzer.semver:semver:${PluginVersions.semver}")
+    implementation("org.eclipse.jgit:org.eclipse.jgit:${PluginVersions.jgit}")
```
JLLeitschuh: Does this drag along any transitive dependencies?

JLLeitschuh:
Why not just simplify this whole process by allowing a user to specify a custom `RegularDirectoryProperty` (or whatver the type is) that declares a custom root?

plugin/build.gradle.kts
Tapchicoma: It brings them:
```
\--- org.eclipse.jgit:org.eclipse.jgit:5.6.0.201912101111-r
     +--- com.jcraft:jsch:0.1.55
     +--- com.jcraft:jzlib:1.1.1
     +--- com.googlecode.javaewah:JavaEWAH:1.1.6
     +--- org.slf4j:slf4j-api:1.7.2
     +--- org.bouncycastle:bcpg-jdk15on:1.64
     |    \--- org.bouncycastle:bcprov-jdk15on:1.64
     +--- org.bouncycastle:bcprov-jdk15on:1.64
     \--- org.bouncycastle:bcpkix-jdk15on:1.64 (*)
```

Tapchicoma:
> Why not just simplify this whole process by allowing a user to specify a custom RegularDirectoryProperty (or whatver the type is) that declares a custom root?

Using JGit make it a little more automatic and require less configuration from the user. As directory could not be in the multiple git repos at the same time, I think it is ok to have such automation.

@JLLeitschuh Initial implementation was naive, I've did it properly in force-updated  commit, could you look at it again?

plugin/build.gradle.kts
JLLeitschuh: These seems like quite a few dependencies to drag along with the plugin.
If you really think this is necessary, ideally we'd vendor them inside of our own JAR file instead.

Tapchicoma: > If you really think this is necessary, ideally we'd vendor them inside of our own JAR file instead.

Do you mean creating shadowed fat-jar gradle plugin?

JLLeitschuh: More or less , that's the idea. Something that re-writes the imports to be something we vendor.

Tapchicoma: I am not a big fan of vendoring in java world as it becomes consumer much harder (ask publisher to do a new release) to update 3rd party dependencies if there are some urgent bugfix

JLLeitschuh: I can't find the documentation currently, but it was a common recommendation to try not to bring along dependencies in plugins. Primarily so you significantly reduce the chances that two plugins drag along conflicting dependencies. 

That's quite a few dependencies to be dragging along especially given that we're using such a minor bit of the libraries functionality. If you really, really need to use this dependency, I advise that instead you create a custom configuration so it's loaded on an isolated classpath and call that method via reflection.

Tapchicoma: You convinced me :slightly_smiling_face: 

I will implement fat jar and put there this dependencies as well as semver dependency.

JLLeitschuh: That works. Thanks!

Tapchicoma: Done in https://github.com/JLLeitschuh/ktlint-gradle/pull/323/commits/07a26c08362fac95ec5d3037b8c4f80112bab684

It took me a while to combine shadow, java-gradle-plugin and com.gradle.plugin-publish together :disappointed_relieved: 

Happy to hear possible integration improvements.

plugin/build.gradle.kts
```diff
@@ -82,6 +129,27 @@ gradlePlugin {
...
+afterEvaluate {
+    publishing {
+        publications {
+            withType(MavenPublication::class.java) {
+                // Special workaround to publish shadow jar instead of normal one. Name to override peeked here:
+                // https://github.com/gradle/gradle/blob/master/subprojects/plugin-development/src/main/java/org/gradle/plugin/devel/plugins/MavenPluginPublishPlugin.java#L73
+                if (name == "pluginMaven") {
+                    setArtifacts(
+                        listOf(
+                            project.tasks.named("shadowJar").get()
```
JLLeitschuh: Don't you already have the `shadowJarTask` in scope?

Tapchicoma: Good point, will use it :+1:

Tapchicoma: Updated

plugin/build.gradle.kts
```diff
@@ -82,6 +131,27 @@ gradlePlugin {
...
+                    setArtifacts(
+                        listOf(
+                            shadowJarTask.get()
+                        )
+                    )
+                }
+            }
+        }
+    }
+}
```
JLLeitschuh: Can you verify that the `pom.xml` that we're generating doesn't have these shadowed dependencies?

If there's a way to verify with a build task that we aren't adding any unexpected additional dependencies, that would be useful.

Also, something like this may help us out too for sanity checking.
```
val ensureDependenciesAreInlined by tasks.registering {
    dependsOn(tasks.shadowJar)
    doLast {
        zipTree(tasks.shadowJar.flatMap { it.archiveFile }).visit {
            if (!isDirectory) {
                val path = relativePath
                if (path.lastName.endsWith(".class") && !path.pathString.startsWith(pluginGroup.replace(".", "/")) {
                    throw GradleException("Found non inlined dependency: " + path.pathString)
                }
            }
        }
    }
}
```

Tapchicoma: By running `publishToMavenLocal` I see that neither `pom` file nor gradle metadata do not contain dependencies. 

That was the reason I've introduced `shadowImplementation` configuration - is to remove `implementation` dependencies to be added to generated pom/gradle metadata, but still shadow them into final jar file.

I've also tried to update to snapshot version in my work project - works ok.

You task looks interesting, will see if it works :+1:

Tapchicoma: I've added proposed task in https://github.com/JLLeitschuh/ktlint-gradle/pull/323/commits/c56aa4551e027407a1b38220d917de4e6389b5c1 commit

PR #325, opened 2020-01-16
Add `pull_request` to Wrapper Validation Triggers

JLLeitschuh:
There was a bug. Whoops.


JLLeitschuh:
There will be two validate checks for maintainers, only one for external contributors.

PR #333, opened 2020-02-06
Integrate "com.github.breadmoirai.github-release" plugin.

Tapchicoma:
This plugin will automatically sync changelog to github releases page on a new plugin release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,11 @@ script:
   - ./plugin/gradlew -p ./plugin ktlintCheck --no-daemon
   - ./gradlew ktlintCheck --no-daemon
   - ./gradlew addKtlintCheckGitPreCommitHook --no-daemon && checkbashisms .git/hooks/pre-commit
-  - if [ -n "$TRAVIS_TAG" ]; then ./plugin/gradlew -p ./plugin publishPlugins --no-daemon; fi
+  - |
+      if [ -n "$TRAVIS_TAG" ]; then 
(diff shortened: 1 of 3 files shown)


Tapchicoma:
@JLLeitschuh you will have to add `GITHUB_KEY` to Travis secrets. Key value could be retrieved from "user settings > developer settings" according to https://github.com/BreadMoirai/github-release-gradle-plugin#using-this-plugin

JLLeitschuh:
Okay token has been added.

.travis.yml
```diff
@@ -17,6 +17,7 @@ script:
   - ./gradlew ktlintCheck --no-daemon
   - ./gradlew addKtlintCheckGitPreCommitHook --no-daemon && checkbashisms .git/hooks/pre-commit
   - if [ -n "$TRAVIS_TAG" ]; then ./plugin/gradlew -p ./plugin publishPlugins --no-daemon; fi
+  - if [ -n "$TRAVIS_TAG" ]; then ./plugin/gradlew -p ./plugin githubRelease --no-daemon; fi
```
JLLeitschuh: Why not run both tasks in the same build?

Tapchicoma:
> This LGTM. Have you verified the logic on a personal repository?
Nope, never used it before. Will see how it works with next release :sweat_smile:

.travis.yml
Tapchicoma: Updated configuration to use same if condition.

PR #343, opened 2020-02-16
Don't include files into commit on running check hook.

### LINKED ISSUE #330: The pre-commit Git hook adds files after checking them
This line of the pre-commit hook causes problems for me: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt#L46

I frequently make commits that contain just hunks of a file using `git add --patch` (usually when I have several changes locally that I want to split into separate commits).  When the pre-commit runs on files where I've done that, it blindly stages the rest of the file, changing what I was committing.  

My guess is that this is needed for the `ktlintFormat` pre-commit hook to add any changes made into the commit.  However, I believe this is not needed for the `ktlintCheck` pre-commit hook (which is what I'm using), since it does not change any files.

I've fixed this for myself by simply removing that loop from the hook.

Tapchicoma:
Format one will still do it.

Closes #330


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -86,6 +95,7 @@ private fun KtlintPlugin.PluginHolder.addInstallGitHookFormatTask() {
         it.description = "Adds git hook to run ktlintFormat on changed files"
         it.group = HELP_GROUP
         it.taskName.set(FORMAT_PARENT_TASK_NAME)
+        it.shouldUpdateCommit.set(true)
```
JLLeitschuh: I think this is appropriate?
```suggestion
        it.shouldUpdateCommit.convention(true)
```

Tapchicoma: I would say no. As I see from documentation - `.convention(value)` should be used to set default value of `Property`. In this line value is pretty specific to this task.

JLLeitschuh: Perhaps we should set the convention inside of the task?

Tapchicoma: that will make sense :+1:

Tapchicoma: @JLLeitschuh I've set convention on task property (default is `false`) and forced-pushed.

JLLeitschuh: Why is the convention false but this is set to `true` here?

Tapchicoma: I've set convention in task class `KtlintInstallGitHookTask`. In this particular lines specific task "addKtlintFormatGitPreCommitHook" is registered and `shouldUpdateCommit` changed to `true` as hook will run `ktlint --format` over staged files.

JLLeitschuh: Can you leave that as a comment to explain that to the next person that sees this?

Tapchicoma: Added a comment in updated commit https://github.com/JLLeitschuh/ktlint-gradle/pull/343/commits/43fabe7bbc8c649266a0b7c2c4a4a5ccd7fe4aa2

PR #346, opened 2020-03-07
Allow to override generated reports base directory.

### LINKED ISSUE #321: Exposing output file for reporters in plugin ktlint properties
We would like to have the option to choose the location for reporter, which in ktlint is optional, but in plugin it's hardcoded and not available for setting.

```
        project.layout.buildDirectory.file(project.provider {
            "reports/ktlint/${this@BaseKtlintCheckTask.name}.$fileExtension"
        })
```

Tapchicoma:
Closes #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ The assumption being that you would not want to lint code you weren't compiling.
     - [Simple setup](#idea-plugin-simple-setup)
     - [Using new plugin API](#idea-plugin-setup-using-new-plugin-api)
   - [Plugin configuration](#configuration)
+    - [Setting reports output directory](#setting-reports-output-directory)
     - [Customer reporters](#custom-reporters)
   - [Samples](#samples)
 - [Task details](#tasks-added)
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
```diff
@@ -32,13 +33,34 @@ import org.jlleitschuh.gradle.ktlint.reporter.ReporterType
...
+     */
+    @get:Internal
+    val reporterOutputDir: DirectoryProperty = objectFactory.directoryProperty().convention(
+        project.layout.buildDirectory.dir("reports/ktlint")
+    )
+
+    /**
+     * [DirectoryProperty] takes into account also content of directory, when snapshotting input.
+     * As reports generated are always generated in this directory, this prevents task proper caching,
+     * so this input exists as a workaround for the described problem.
```
JLLeitschuh: Can you elaborate? I'm missing something.

JLLeitschuh: Task caching also snapshots the outputs. If the output has changed, the task is marked as out of date. Why do we need this?

Tapchicoma: You are right, was confused by https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputDirectory.html. Actually `reporterOutputDir` could be just annotated with `@get:Input`.

Tapchicoma: Force updated and removed this property.

Tapchicoma: or not... running tests  in IDEA was green, though from command line they fail :thinking:

Tapchicoma: Should be fine now

README.md
```diff
@@ -282,6 +283,32 @@ dependencies {
...
 
+#### Setting reports output directory
+
+It is possible also to define different from default output directory for generated reports
+ (by default it is "build/reports/ktlint"):
+<details>
+<summary>Groovy</summary>
+
+```groovy
+tasks.withType(org.jlleitschuh.gradle.ktlint.KtlintCheckTask.class) {
```
JLLeitschuh, line 294:
```suggestion
tasks.withType(org.jlleitschuh.gradle.ktlint.KtlintCheckTask) {
```

In groovy, I don't think you need the `.class`

PR #352, opened 2020-04-23
Gradle Instant Execution: invocation of 'Task.project' at execution time is unsupported

CristianGM:
Running a project with the flag `-Dorg.gradle.unsafe.instant-execution=true` results in a failure because:
> task `:ktlintMainSourceSetCheck` of type `org.jlleitschuh.gradle.ktlint.KtlintCheckTask`: invocation of 'Task.project' at execution time is unsupported.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -18,7 +18,7 @@ open class KtlintCheckTask @Inject constructor(
 
     @TaskAction
     fun lint(inputChanges: InputChanges) {
-        project.logger.info("Executing ${if (inputChanges.isIncremental) "incrementally" else "non-incrementally"}")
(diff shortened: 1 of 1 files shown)


eskatos:
Note that there is more code in this plugin that access `Project` at execution time (e.g. `project.projectDir` that should be replaced by `ProjectLayout` or `project.javaexec {}` that should be replaced by `ExecOperations`.

The Kotlin Gradle Plugin also having some offending code currently gets in the way of making `ktlint-gradle` tests pass with more fixes and instant execution enabled. I'll revisit that when it's time.

CristianGM:
@eskatos but there is no warning about it (from a consumer point of view)

eskatos:
@CristianGM it depends on which tasks from the plugin are in the task graph. I got more problems reported by running the whole plugin test suite modified to enable instant execution, running against the latest Gradle nightly.

I'll push an experimentation branch on my fork.

eskatos:
Pushed at https://github.com/eskatos/ktlint-gradle/pull/1

Issue #350, opened 2020-04-11
Set verification group on ktlint tasks


ZacSweers:
Currently tasks are group-less, which means they go on to a "other" catch-all. Would be nice if these could go in the verification group. Happy to send a PR.

ZacSweers:
Any thoughts?

ZacSweers:
Could we put them in a ktlint group then?

Tapchicoma:
> Gradle doesn't support marking a task as "private". However, gradle tasks will only show a task if it either has task.group set, or no other task depends on it.

Actually we probably could set verification group on all tasks :thinking: Source Set tasks should be hidden as meta tasks depend on them. Could you create a PR?

JLLeitschuh:
What's the use case here again? Sorry.

ZacSweers:
To group them appropriately rather than all get dumped in "other"

JLLeitschuh:
From what I've seen, this plugin can end up generating quite a few tasks. Especially when you have the cross compilation Kotlin plugins applied. I'm concerned that a change like this adds little value and will just result in information overload in the output of `:tasks`.

There's also the argument that many of the tasks that we generate are 'internal' and should only be executed in aggregate. Maybe I'm wrong.

When you add the 'group' it usually indicates to the user that we're fine that you explicitly invoke those specific sub-tasks instead of running the aggregator tasks. I don't know if that's the kind of behaviour we actually want to be encouraging.

Thoughts?

PR #357, opened 2020-05-11
Specify group in BaseKtlintCheckTask

ZacSweers:
Resolves #350

PR #358, opened 2020-05-11
Specify group for KtlintApplyToIdeaTask

ZacSweers:
Part of #350


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintApplyToIdeaTask.kt
```diff
@@ -14,6 +14,10 @@ import org.gradle.api.tasks.TaskAction
 open class KtlintApplyToIdeaTask @Inject constructor(
     objectFactory: ObjectFactory
 ) : DefaultTask() {
+
```
ZacSweers, line 17:
```suggestion

```

ZacSweers:
Merging into #357

PR #361, opened 2020-06-04
Set default ktlint version to 0.37.1

friederbluemle:
- Use ktlint `0.37.1` by default, which required minor formatting fixes in the android sample app
- Patch updated the Android Gradle plugin from `3.6.0` to `3.6.3` (including the bundled kotlin dependency)
- A couple of minor typo fixes (in a separate commit).


Tapchicoma:
@friederbluemle thank you for your contribution. Could you update ktlint version to `0.37.1` and fix few codestyle issues in samples? (see failing CI check).

knyttl:
can we pls merge this? also ktlint is now 0.37.2

Tapchicoma:
@knyttl you could always set specific ktlint version to use via plugin extension configuration. `0.37.2` should work without problems.

PR #386, opened 2020-08-26
Skip ktlint and report up-to-date if incremental task run contains only removed files

### LINKED ISSUE #385: Incremental check may incorrectly run against all source files instead of the correct source set
Versions:
ktlint-gradle: 9.1.1
Gradle version: 6.1.1
Kotlin version: 1.3.61

The `Check` task [filters out](https://github.com/JLLeitschuh/ktlint-gradle/blob/fa5614c7e5652bb7f8726da4c43705d165517e57/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt#L28) incremental changes of type `DIRECTORY` or `REMOVED`. If the only changes to a source set as reported by Gradle have those types, then the resulting set of source files after the filter is empty. This causes `ktlint` to be invoked with an args file that does not list any source files to check. In that case, the default behavior of `ktlint` is to check ALL source files in the current directory.

I believe the correct behavior in the case where Gradle reports incremental changes, but all such changes are filtered out, would be to set the `didWork` property to false and return without executing `ktlint`.

I will try to create a repro and possibly submit a PR when I have time.

davidnelson-hms:
Fixed #385 

Test and main code are in two separate commits to make it easier to demonstrate the problem. I would suggest squashing at merge.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintCheckTask.kt
@@ -31,8 +31,14 @@ open class KtlintCheckTask @Inject constructor(
             }
             .map { it.file }
             .toSet()
-        logger.debug("Files changed: $filesToLint")
 
-        runLint(filesToLint)
+        if (filesToLint.isEmpty()) {
+            didWork = false
+            logger.info("No ${ChangeType.ADDED} or ${ChangeType.MODIFIED} files that need to be linted")
+        } else {
+            logger.debug("Files changed: $filesToLint")
+
+            runLint(filesToLint)
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


davidnelson-hms:
Hilariously, I didn't realize when I first submitted this that the `plugin` project is an _included_ build, not a subproject, so running `build` at the root doesn't fully build the `plugin` project (my bad for not fully reading the README; although now I am curious at the motivation for the project structure). So I didn't notice that my changes were failing the build because of...lint failures...

(fixed now)

Tapchicoma:
Plugin is included build so samples always uses the current state of the plugin without any publishing somewhere. Probably now there is exists a better approach.

Tapchicoma (approved):
Thank you for fixing it :+1:

Tapchicoma:
@davidnelson-hms could you fix codestyle issues?

PR #389, opened 2020-09-02
Fix tasks may not be cacheable due to overlapping outputs.

### LINKED ISSUE #379: KtlintCheckTask not cacheable due to overlapping outputs
From my build scan:
> Overlapping outputs: Gradle does not know how file 'build\reports\ktlint\ktlintAndroidTestSourceSetCheck.txt' was created (output property 'reporterOutputDir'). Task output caching requires exclusive access to output paths to guarantee correctness.

I have several `KtlintCheckTask` tasks in my build, and all are noncacheable for this reason. The tasks run every time even when Gradle runs with `--build-cache`.

ktlint-gradle version 9.3.0

Tapchicoma:
Fixes #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   - ?
 ### Changed
   - Updated Gradle to `6.6.1` version
+  - Each task will output reports into subdirectory inside `build/reports/ktlint` directory to fix non-working caching [#379](https://github.com/JLLeitschuh/ktlint-gradle/issues/379)
 ### Fixed
   - ?
 ### Deleted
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   - ?
 ### Changed
   - Updated Gradle to `6.6.1` version
+  - Each task will output reports into subdirectory inside `build/reports/ktlint` directory to fix non-working caching [#379](https://github.com/JLLeitschuh/ktlint-gradle/issues/379)
```
JLLeitschuh, line 13: The output file location is technically an API breaking change, if users are relying directly upon the output location, but from a compile standpoint it isn't.

What are your thoughts? Is this worthy of a major version bump?

Tapchicoma: I think it ok to be minor release. Those who want old "behaviour" could redefine it to old approach.

PR #400, opened 2020-09-30
Fix running checks on windows does not fail on errors.

### LINKED ISSUE #232: Run Windows CI Jobs
Would be good to integrate Windows CI jobs into our pipeline to sanity check that everything still works regardless of OS.

https://azure.microsoft.com/en-us/services/devops/pipelines/

### LINKED ISSUE #399: ktlintCheck task does not fail on Windows
Hello Ktlint-Gradle team.

I'm facing the strange issue.
When I'm running the local ktlint check via `./gradlew clean ktlintCheck` the _"mainSourceSetCheck"_ task completes successfully.
But when I'm running the same process in `docker build` environment the result is different and it fails.

I've found the same problem on StackOverflow: 
[https://stackoverflow.com/questions/63863368/ktlint-check-doesnt-work-locally-anymore-it-works-tho-in-my-docker-image-buil](https://stackoverflow.com/questions/63863368/ktlint-check-doesnt-work-locally-anymore-it-works-tho-in-my-docker-image-buil)

Versions I'm using:
 - Gradle wrapper 6.6.1-bin.zip
 - OpenJDK 14.0.2
 - Kotlin version 1.4.10
 - Ktlint-gradle plugin 9.4.0

The `java --version` output in the local and docker environment is the same:
```
openjdk 14.0.2 2020-07-14 
OpenJDK Runtime Environment (build 14.0.2+12-46) 
OpenJDK 64-Bit Server VM (build 14.0.2+12-46, mixed mode, sharing)
```

Local environment check outputs the following message `[DEBUG] <...>ms / 0 file(s) / 0 error(s)`
whereas the docker environment outputs the `[DEBUG] <...>ms / 1 file(s) / 1 error(s)`

I've also made a minimum reproducible example. You may found it here: 
[https://github.com/santelos/ktlint-mre](https://github.com/santelos/ktlint-mre)

---

build.gradle.kts
```
repositories {
    mavenCentral()
}

plugins {
    kotlin("jvm") version "1.4.10"
    id("org.jlleitschuh.gradle.ktlint") version "9.4.0"
}

ktlint {
    debug.set(true)
}

tasks.withType<KotlinCompile> {
    kotlinOptions.jvmTarget = "14"
}
```

Dockerfile
```
FROM openjdk:14.0.2
WORKDIR /build
COPY . .
RUN java --version
RUN ./gradlew clean ktlintCheck
```

Tapchicoma:
- Enable running plugin tests on windows os via Github actions
- Fix Picoli [requires](https://picocli.info/#AtFiles) double `\\` in argument files
- Fix failing tests

Fixes #399 
Fixes #232

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -4,7 +4,10 @@ on: [pull_request, push]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v1
       - name: set up JDK 1.8
@@ -59,7 +62,10 @@ jobs:
         run: ./gradlew addKtlintCheckGitPreCommitHook --no-daemon && checkbashisms .git/hooks/pre-commit
 
   check_samples:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     needs: build
     steps:
       - uses: actions/checkout@v1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [9.4.1] - 2020.10.05
+### Fixed
+  - Plugin now correctly validates files on Windows OS [#399](https://github.com/JLLeitschuh/ktlint-gradle/issues/399)
+
 ## [9.4.0] - 2020.09.06
 ### Changed
   - Updated Gradle to `6.6.1` version
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
@@ -207,7 +207,8 @@ abstract class BaseKtlintCheckTask(
                 }
             additionalConfig(argsWriter)
             filesToCheck.forEach {
-                argsWriter.println("\"${it.toRelativeFile()}\"")
+                val filePath = it.toRelativeFile().path.replace("\\", "\\\\")
+                argsWriter.println("\"$filePath\"")
             }
         }
         return argsConfigFile
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
@@ -242,7 +242,7 @@ abstract class BaseKtlintCheckTask(
         ) {
             logger.warn(
                 "You are using ktlint version ${ktlintVersion.get()} that has the security vulnerability " +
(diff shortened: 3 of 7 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -404,11 +404,18 @@ abstract class BaseKtlintPluginTest : AbstractPluginTest() {
             """.trimIndent()
         )
 
+        // TODO: Add better approach to different file paths on different OS
```
Tapchicoma, line 407: Most probably it may go away with upcoming plugin changes.

.github/workflows/build-and-check.yml
```diff
@@ -4,7 +4,10 @@ on: [pull_request, push]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
JLLeitschuh, line 9: Thank you for doing this. This was going to be my first request.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [9.4.1] - 2020.10.05
+### Fixed
+  - Plugin may not validate files on Windows OS and report SUCCESS [#399](https://github.com/JLLeitschuh/ktlint-gradle/issues/399)
```
JLLeitschuh, line 10: Shouldn't this be in the present tense, instead of the past tense. IE:
"Plugin now correctly validates files on Windows OS and reports SUCCESS"

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
```diff
@@ -207,7 +207,8 @@ abstract class BaseKtlintCheckTask(
                 }
             additionalConfig(argsWriter)
             filesToCheck.forEach {
-                argsWriter.println("\"${it.toRelativeFile()}\"")
+                val filePath = it.toRelativeFile().path.replace("\\", "\\\\")
```
JLLeitschuh, line 210: Shouldn't this be `replaceAll`? Doesn't `replace` only replace the first occurance?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -404,11 +404,18 @@ abstract class BaseKtlintPluginTest : AbstractPluginTest() {
...
+        fun String.updatePathForWindows() = if (System.getProperty("os.name").contains("Windows")) {
+            replace("/", "\\\\")
```
JLLeitschuh, line 409: Same question as above about `replaceAll` vs `replace`.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/BaseKtlintCheckTask.kt
Tapchicoma: Kotlin does not have `replaceAll`. `replace` function replaces all occurrences.

CHANGELOG.md
Tapchicoma: Force-pushed changelog update

JLLeitschuh (approved):
Nice work! Thanks for improving our test coverage to cover windows!! 😄

PR #403, opened 2020-10-11
Update Kotlin to 1.4.10 version.

Tapchicoma:
Drop support for deprecated Kotlin plugins.

First of planned breaking changes for next `10.0.0` release.


plugin/build.gradle.kts
```diff
@@ -21,6 +22,12 @@ repositories {
     jcenter()
 }
 
+tasks.withType<KotlinCompile>() {
+    kotlinOptions {
+        apiVersion = "1.3"
```
JLLeitschuh, line 27: Hrm... Gradle is still pretty heavily reliant upon version 1.2. Correct? If we start using 1.3 features, we'll break backwards compatibility with older versions of Gradle, won't we?

Tapchicoma: apiVersion `1.3` was default target for the code compiled with Kotlin version `1.3.*`. Also Gradle itself is using Kotlin `1.3.*` for kotlin scripts, so I assume `1.3` here is correct.

Also samples works correctly.

JLLeitschuh: Oh, I was thinking about Kotlin 1.4
Nevermind. Thanks!

JLLeitschuh (approved):
LGTM!

PR #409, opened 2020-11-01
Migrating publishing to GitHub actions.

Tapchicoma:


.github/workflows/new-release.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          path: ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles('**/gradlew') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-wrapper-
+      - name: Publish plugin
+        env:
+          GRADLE_PUBLISH_KEY: {{ secrets.GradlePublishKey }}
+          GRADLE_PUBLISH_SECRET: {{ secrets.GradlePublishSecret }}
+          GITHUB_KEY: {{ secrets.GithubKey }}
```
Tapchicoma, line 36: @JLLeitschuh please check here secrets names

JLLeitschuh: Should be `GRADLE_PUBLISH_KEY` and `GRADLE_PUBLISH_SECRET`. I never setup a GitHub key. Isn't there one that's implicit in the environment?

Tapchicoma: I don't see anything related to `GITHUB_KEY` here: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

Tapchicoma: I've used proper secret name for `GRADLE_PUBLISH_KEY` and `GRADLE_PUBLISH_SECRET` via force-push.

JLLeitschuh (approved):
LGTM!

PR #411, opened 2020-11-11
Update Android Gradle plugin to 4.1 version.

Tapchicoma:
Breaking change as I used new api and dropped support for build variants meta tasks. New api does not provide such information.


samples/kotlin-mpp-android/build.gradle.kts
```diff
@@ -8,17 +8,8 @@ kotlin {
...
-            }
-        }
-
-        val androidMain by getting {
-            dependencies {
-                implementation(kotlin("stdlib"))
-            }
-        }
+        val commonMain by getting
+        val androidMain by getting
```
JLLeitschuh, line 12: Are these still needed at all?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt
```diff
@@ -0,0 +1,108 @@
+@file:Suppress("UnstableApiUsage")
```
JLLeitschuh, line 1: Can you apply this a bit more specifically?

JLLeitschuh (changes requested):
A few small things.

Tapchicoma:
@JLLeitschuh sorry for not addressing your review for a long time. Just did it, please check fixes.

JLLeitschuh (approved):
LGTM Thanks!

PR #421, opened 2020-12-20
Update ktlint to 0.40.0.

utwyko:
See https://github.com/pinterest/ktlint/releases/tag/0.40.0


JLLeitschuh:
Looks like this change introduces some test failures. PTAL

utwyko:
> Looks like this change introduces some test failures. PTAL

Huh seems I ran the tests incorrectly locally before pushing. I'll take a look later this week!

utwyko:
Tests are failing due to https://github.com/pinterest/ktlint/issues/997. Will update this PR when a version of ktlint is released including that fix.

Tapchicoma:
Probably better to add empty `.editorconfig` then to wait for a new release.

Issue #229, opened 2019-04-23
Always fail task on Ktlint crash


tasomaniac:
I've noticed that while running any of the `KtlintCheck` tasks, if running ktlint executable fails, the task is not marked as failure. 

Here is an example with incompatibility with Kotlin 1.3.30

```

> Task :project:ktlintMainSourceSetCheck
Exception in thread "main" java.lang.NoSuchFieldError: PROPERTY_DELEGATE
	at com.github.shyiko.ktlint.core.ast.ElementType.<clinit>(ElementType.kt:33)
	at com.github.shyiko.ktlint.ruleset.standard.ChainWrappingRule.<init>(ChainWrappingRule.kt:36)
	at com.github.shyiko.ktlint.ruleset.standard.StandardRuleSetProvider.get(StandardRuleSetProvider.kt:12)
	at com.github.shyiko.ktlint.Main.main(Main.kt:310)
```

The exception is recorded, ktlint actually does not run, but the task is marked as success.

Tapchicoma:
@tasomaniac could it happen that you have enabled `ignoreFailuers` in `ktlint-gradle` extension?

```kotlin
val ignoreFailures: Property<Boolean> = objectFactory.property { set(false) }
```

tasomaniac:
Hmm actually I did and then I have another task of myself that evaluates the `xml` reports.

I wonder if there is a way to differentiate `failures` and `errors` though. I would like to ignore failures resulting from a successful run of the ktlint task. But not for errors that causes `ktlint` task to crash and never run. 

I'm not sure if this is technically possible but would be good to differentiate these two scenarios.

Tapchicoma:
I doubt that it will be possible, but will try to investigate.

tasomaniac:
I can also have a look if you agree that it's a good idea

Tapchicoma:
Feel free to do it. I think it is a good idea to fail the build on ktlint crash when `ignoreFailures` is enabled.

JLLeitschuh:
Is this still an issue? Has it been resolved?

tasomaniac:
I believe this is technically still an issue. But not sure how important it is. 

One way to solve this would be to create an xml report and include the execution issues as errors there. I think Android Lint does that.

JLLeitschuh:
Are you able to create a demo or, even better, a unit test that demonstrates this issue?

PR #424, opened 2021-01-03
Migrate plugin to use KtLint directly instead invoking it via CLI.

### LINKED ISSUE #395: Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
I get deprecated features warning. I used Gradle version is 6.5

./gradlew ktlintFormat
```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings
```

./gradlew ktlintFormat --warning-mode all
```
Querying the mapped value of task ':ktlintKotlinScriptFormat' property 'reporterOutputDir' before task ':ktlintKotlinScriptFormat' has completed has been deprecated. This will fail with an error in Gradle 7.0. Consult the upgrading guide for further information: https://docs.gradle.org/6.5/userguide/upgrading_version_6.html#querying_a_mapped_output_property_of_a_task_before_the_task_has_completed
Querying the mapped value of task ':ktlintMainSourceSetFormat' property 'reporterOutputDir' before task ':ktlintMainSourceSetFormat' has completed has been deprecated. This will fail with an error in Gradle 7.0. Consult the upgrading guide for further information: https://docs.gradle.org/6.5/userguide/upgrading_version_6.html#querying_a_mapped_output_property_of_a_task_before_the_task_has_completed
Querying the mapped value of task ':ktlintTestSourceSetFormat' property 'reporterOutputDir' before task ':ktlintTestSourceSetFormat' has completed has been deprecated. This will fail with an error in Gradle 7.0. Consult the upgrading guide for further information: https://docs.gradle.org/6.5/userguide/upgrading_version_6.html#querying_a_mapped_output_property_of_a_task_before_the_task_has_completed
```

Tapchicoma:
Main motivation of this huge refactoring is to migrate plugin to use directly `KtLint` class instead of doing `javaexec` for CLI. 

Several reasons are behind this:
- Gain more control over how to run ktlint and how to output errors
- CLI creates thread pool itself to run checks in parallel. This thread pool competes over CPU resources with Gradle daemon threads, when plugin is used in multimodule project and `org.gradle.parallel` flag is enabled.
- Use more optimized approach to preload info and run lint checks itself (more about it below)

This PR introduces new tasks and task hierarchy, plus uses workers everywhere. `WorkAction`s are running in isolated classloader mode, allowing to still support seamless KtLint version changes without updating plugin itself (unless KtLint will have some breaking changes). 

Following task structure is created:
- ~`LoadRuleSetsTask` - preloads `RuleSetProvider`s and serialize them into intermediate file. Such approach allows to skip this step on incremental run and start doing linting directly.~ - After some testing, I see this task adds no performance gain and I removed it.
- `LoadReportersTask` - preloads `ReporterProvider` and based on enabled and custom reporters serialize two files - one with only enabled `ReporterProviders`, second with some useful info for `GenerateReportsTask`. Again such approach allows to skip this step on incremental run and start doing linting directly.
- `BaseKtlintCheckTask`* runs as previously actual linting. Currently only one `WorkAction` is submitted for all files in `SourceSet`.
- `GenerateReportsTask` - generates actual reports and writes errors into Gradle console. This task will fail the build if errors were not autocorrected and present. Each report generation is running in own `WorkAction`, plus console output is additional `WorkAction`.

Added a tests to verify supported KtLint versions are working correctly and they are besides two:
- `0.37.0` on Windows crashes. This was fixed in `0.37.1` patch release.
- `0.38.0` on all OS crashes, because it was build with Kotlin api 1.4 level that is not supported yet by Gradle. Issue was fixed in `0.38.1` patch release.

Did some testing on average project (312 modules, 4150 Kotlin files, 240_000 lines of code):
- current ktlint-gradle release has around 7 min runtime to check all
- snapshot release from this PR has around 2.5 min runtime to check all

Single `SourceSet` check runtime is slightly increased now. For example, for module with 232 Kotlin files it is increased from 10.1s to 17.4s. This single module performance issue could be addressed by splitting files into buckets in followup PR.

Changes itself fix #395 and #229 issues.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
```diff
@@ -93,14 +93,13 @@ internal const val VERIFICATION_GROUP = LifecycleBasePlugin.VERIFICATION_GROUP
 internal const val FORMATTING_GROUP = "Formatting"
 internal const val HELP_GROUP = HelpTasksPlugin.HELP_GROUP
 internal const val CHECK_PARENT_TASK_NAME = "ktlintCheck"
-internal const val FORMAT_PARENT_TASK_NAME = "ktlintFormat"
+internal const val FORMAT_PARENT_TASK_NAME = "kllintFormat"
```
JLLeitschuh, line 96: Spelling mistake here

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/CustomReporter.kt
```diff
@@ -1,41 +1,26 @@
...
-    internal val dependencyArtifact: Dependency get() {
-        val artifact = _dependencyArtifact
-        requireNotNull(artifact) { "Reporter dependency is not set!" }
-        return artifact
+    val reporterId: String = name,
+    var fileExtension: String = reporterId,
+    @Transient var dependency: Any? = null,
+) : Serializable {
+    companion object {
+        private const val serialVersionUID: Long = 2012775L
```
JLLeitschuh, line 24: Why is this now serializable?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/ConsoleReportWorkAction.kt
```diff
@@ -0,0 +1,67 @@
...
+                    .forEach {
+                        it.first.lintError.logError(filePath, verbose)
+                    }
+            }
+        }
+
+        if (errors.any { it.lintErrors.any { !it.second } } &&
+            !parameters.ignoreFailures.getOrElse(false)
+        ) {
+            throw GradleException("KtLint found code style violations.")
```
JLLeitschuh, line 37: In the future, or now, including the path to the report would be nice here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/GenerateReportsWorkAction.kt
```diff
@@ -0,0 +1,66 @@
...
+                it.readObject() as List<SerializableReporterProvider>
+            }
+            .map { it.reporterProvider }
+
+    private fun loadErrors(): List<LintErrorResult> =
+        ObjectInputStream(FileInputStream(parameters.discoveredErrorsFile.asFile.get()))
+            .use {
+                @Suppress("UNCHECKED_CAST")
+                it.readObject() as List<LintErrorResult>
+            }
```
JLLeitschuh, line 57: Why are we shipping around serialized java objects here? This doesn't seem like the best format, does it?

JLLeitschuh: Also, if we do chose to go with `ObjectInputStream`, we need to use something like the Apache commons `ValidatingObjectInputStream` to prevent the deserialization of arbitrary java objects.

JLLeitschuh (changes requested):
Not sure this is headed in the right direction, lets chat 1:1. See my DM in slack.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/CustomReporter.kt
Tapchicoma: To object needs to implement `Serializable` interface to be able to pass as input parameter to the worker.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/GenerateReportsWorkAction.kt
Tapchicoma: > Why are we shipping around serialized java objects here? This doesn't seem like the best format, does it?

Didn't want to bring any additional dependency to do serialization/deserialization. Do you have something in mind?

> Also, if we do chose to go with ObjectInputStream, we need to use something like the Apache commons ValidatingObjectInputStream to prevent the deserialization of arbitrary java objects.

Will check it :eyes:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt
Tapchicoma: Fixed via 2145957

pbielicki:
@Tapchicoma I see this PR is in `Draft` mode - when do you plan to finalize all the necessary changes?
Can you pls upgrade to the latest Gradle version `6.7.1`? - it has the benefit of showing deprecation warnings in the build scan.

@JLLeitschuh regarding deprecation warnings in our builds coming from ktlint plugin: it would be great to know the timeline for this PR to be merged. If it's still work in progress and it requires several days/weeks we can think of simple workaround e.g. https://github.com/pbielicki/ktlint-gradle/commit/1a914afd72fd13ee2fd6ff97756380cf7fd639eb

Tapchicoma:
@pbielicki there are one issue with disabled rules that I want to fix. Besides that some deserialization security improvements and some additional tests. To not rush - probably start of the next week it will be ready.

> Can you pls upgrade to the latest Gradle version 6.7.1

Will be done in a separate PR before plugin release.

Generally I would expect new release on the next week.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
```diff
@@ -0,0 +1,174 @@
...
+ * Generates reports and prints errors into Gradle console.
+ *
+ * This will actually fail the build in case some non-corrected lint issues.
+ */
+@CacheableTask
+abstract class GenerateReportsTask @Inject constructor(
+    private val workerExecutor: WorkerExecutor,
+    projectLayout: ProjectLayout,
+    objectFactory: ObjectFactory
+) : DefaultTask() {
```
JLLeitschuh, line 39: Sorry, I'm not a huge fan of having this be a separate task. This pattern is in conflict with how most tasks in Gradle operate.

Pretty much all Gradle tasks both do the work and generate the report.

That way, if you invoke the specific task manually from the task selector, you always get the report.

Won't this break that?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/SerializableRuleSet.kt
```diff
@@ -0,0 +1,51 @@
...
+        ruleSet = RuleSet(
+            id,
+            *rules.toTypedArray()
+        )
+    }
+
+    companion object {
+        private const val serialVersionUID: Long = 2012054L
+    }
+}
```
JLLeitschuh, line 51: Not going to lie, seeing hand-crafted object deserialization is making me cringe a bit.

This seems like the perfect place for us to introduce issues when ktlint releases API changing updates.

Tapchicoma: Yes, I plan to add `Serializable` to this objects in next KtLint release. For now we have to use this wrappers.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
Tapchicoma: It was intentional to separate actual linting and generating reports into separate tasks:
- allows in the future to split linting into file buckets that will be handled by workers that run in parallel. This should improve plugin tasks execution time on single module projects with a lot of source files.
- KtLint reports should be generated in one place, so this will prevent splitting linting in buckets.
- Later allow to print all errors from all `SourceSet`s into console in final `ktlintCheck` or `ktlintFormat` tasks
- Probably it will be possible to implement this in the same task, but currently Gradle does not allow to order workers (I am not sure if it is even a good idea to support), so output of one worker treated as input of another one.

> That way, if you invoke the specific task manually from the task selector, you always get the report.

You are wrong here:
- if no input `SourceSet`s are not changed, generate report task will be in `UP_TO_DATE` status
- if input `SourceSet` are changed, generate reports task will still invoke first lint task as generate reports task inputs depends on lint task output.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/GenerateReportsWorkAction.kt
Tapchicoma: Fixed via c1f8260

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/ConsoleReportWorkAction.kt
Tapchicoma: Done via 0182240

pbielicki:
@JLLeitschuh @Tapchicoma how is this going? Any plans when this PR might be merged / released?

README.md
```diff
@@ -339,11 +340,20 @@ This repository provides the following examples of how to set up this plugin:
...
-Additionally, a simple `ktlintCheck` task has also been added that checks all of the source sets for that project.
-Similarly, a `ktlintFormat` task has been added that formats all of the source sets.
-
-Additionally plugin adds two task for project kotlin script files: `ktlintKotlinScriptCheck` and `ktlintKotlinScriptFormat`.
+This plugin adds following tasks to every project:
+- `loadKtlintReporters` - preloads KtLint `Reporter`s
+- `runKtlintCheckOverKotlinScripts` - runs actual lint check over project Kotlin script files
+- `ktlintKotlinScriptCheck` - generates reports and prints issues into Gradle console based on lint check found errors
+- `runKtlintFormatOverKotlinScripts` - tries to format according to the code style project Kotlin script files
+- `ktlintKotlinScriptFormat` - generate reports and prints issues into Gradle console based on found unformattable errors
```
JLLeitschuh, line 348: It still actually attempts to make the formatting change, right? That's not clear here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/KtLintCheckTask.kt
```diff
@@ -38,7 +42,19 @@ open class KtlintCheckTask @Inject constructor(
         } else {
             logger.debug("Files changed: $filesToLint")
 
-            runLint(filesToLint)
+            runLint(filesToLint, false)
         }
     }
+
+    companion object {
```
JLLeitschuh, line 49: Either make this internal, or make the sub methods internal.

```suggestion
    internal companion object {
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/ConsoleReportWorkAction.kt
```diff
@@ -0,0 +1,70 @@
...
+        if (errors.any { it.lintErrors.any { !it.second } } &&
+            !parameters.ignoreFailures.getOrElse(false)
+        ) {
+            val reportsPaths = parameters
+                .generatedReportsPaths
+                .files.joinToString(separator = "\n") { it.absolutePath.prefixIfNot("|- ") }
+
+            throw GradleException(
+                """
+                |KtLint found code style violations. You could find them in following reports:
```
JLLeitschuh, line 43:
```suggestion
                |KtLint found code style violations. Please see the following reports:
```

JLLeitschuh: You don't have to change it to exactly this, but 'could' is past-tense. If you want to keep it as you have it, change it to: 'You can find them'. 'You can find' would be present tense.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        return userData.toMap()
+    }
+
+    private fun loadRuleSetsAndFilterThem(
+        enableExperimental: Boolean,
+        disabledRules: Set<String>
+    ): Set<RuleSet> = loadRuleSetsFromClasspath()
+        .filterKeys { enableExperimental || it != "experimental" }
+        .filterKeys { !(disabledRules.contains("standard") && it == "\u0000standard") }
```
JLLeitschuh, line 97: Why would we get a string starting with the null character here?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
```diff
@@ -0,0 +1,121 @@
...
+        .mapValues { it.value.get() }
+        .values
+        .toSet()
+
+    private fun loadRuleSetsFromClasspath(): Map<String, RuleSetProvider> = ServiceLoader
+        .load(RuleSetProvider::class.java)
+        .associateBy {
+            val key = it.get().id
+            // Adapted from KtLint CLI module
+            if (key == "standard") "\u0000$key" else key
```
JLLeitschuh, line 108: Why is the null character present here?

README.md
Tapchicoma: `ktlintKotlinScriptFormat` depends on `runKtlintFormatOverKotlinScripts` task, so yes, it attempts to make the formatting change.

Will clarify it :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
Tapchicoma: It is copy-paste from KtLint CLI: https://github.com/pinterest/ktlint/blob/aa03a6b6b74cf71d8c2ede0c0a5e2305ce2e4327/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/RuleSetsLoader.kt#L24

Initially it was introduced by Shyiko to enforce "standard" ruleset to be always the first.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
Tapchicoma: Same in the previous comment

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/KtLintCheckTask.kt
Tapchicoma: Fixed via 5ece5b85

README.md
Tapchicoma: Fixed via e4c4017a2

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/ConsoleReportWorkAction.kt
Tapchicoma: Fixed via b5bdc20df

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/LintErrorResult.kt
```diff
@@ -0,0 +1,13 @@
+package org.jlleitschuh.gradle.ktlint.worker
+
+import java.io.File
+import java.io.Serializable
+
+internal data class LintErrorResult(
+    val lintedFile: File,
+    val lintErrors: List<Pair<SerializableLintError, Boolean>>
```
JLLeitschuh, line 8: This depth of generics is kinda a code smell when inside a data class, but I can guess why you're doing it.

Can you at least document what the `Boolean` indicates here in context?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
eskatos: Not sure this would fit the use cases you want to implement but you could consider something like `lintTask.finalizedBy(corresondingReportTask)` so that when one invoke a linting task directly, its companion report task is run afterwards.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
```diff
@@ -0,0 +1,89 @@
...
+        projectLayout.intermediateResultsBuildDir("reporterProviders.bin")
+    )
+
+    @TaskAction
+    fun loadReporters() {
+        checkMinimalSupportedKtLintVersion(ktLintVersion.get())
+
+        val queue = workerExecutor.classLoaderIsolation { spec ->
+            spec.classpath.setFrom(ktLintClasspath, reportersClasspath)
+        }
```
JLLeitschuh, line 66: This doesn't do any dependency collision checking. So you could theoretically end up with multiple instances of, for example, guava here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -0,0 +1,152 @@
...
+            projectLayout.intermediateResultsBuildDir("${name}_errors.bin")
+        )
+
+    protected fun runLint(
+        filesToCheck: Set<File>,
+        formatSources: Boolean,
+    ) {
+        checkDisabledRulesSupportedKtLintVersion()
+
+        val queue = workerExecutor.processIsolation { spec ->
```
JLLeitschuh, line 124: Can you document why you're using process isolation here vs classpath isolation?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
```diff
@@ -0,0 +1,177 @@
...
+                projectLayout
+                    .buildDirectory
+                    .dir("reports${File.separator}ktlint${File.separator}$it")
+            }
+        )
+
+    @Suppress("UnstableApiUsage")
+    @TaskAction
+    fun generateReports() {
+        val queue = workerExecutor.classLoaderIsolation { spec ->
```
JLLeitschuh, line 107: Can you document why you're using classloader isolation here vs process isolation?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
```diff
@@ -0,0 +1,89 @@
...
+    @get:OutputFile
+    internal val loadedReporterProviders: RegularFileProperty = objectFactory.fileProperty().convention(
+        projectLayout.intermediateResultsBuildDir("reporterProviders.bin")
+    )
+
+    @TaskAction
+    fun loadReporters() {
+        checkMinimalSupportedKtLintVersion(ktLintVersion.get())
+
+        val queue = workerExecutor.classLoaderIsolation { spec ->
```
JLLeitschuh, line 64: Can you document why you're using classpath isolation here vs process isolation?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
```diff
@@ -0,0 +1,89 @@
...
+            spec.classpath.setFrom(ktLintClasspath, reportersClasspath)
```
JLLeitschuh, line 65: Why use `setFrom` here instead of `from`? You use `from` in the other two places you call `spec.classpath`

Tapchicoma: No any specific reason, will change it to `from` :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/LintErrorResult.kt
Tapchicoma: Fixed via 3ce901a

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
Tapchicoma: Fixed via 8d561a1

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
Tapchicoma: Fixed via 8d561a1

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
Tapchicoma: Fixed via 8d561a1

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
Tapchicoma: Fixed via 33c8671

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
JLLeitschuh: @Tapchicoma please do do this `finalizedBy`

JLLeitschuh (changes requested):
I think two things:

 1. Add archunt tests to verify that ktlint classes aren't called where they shouldn't be.
 2. Do the `finalizeBy` thing suggested by Paul.

pbielicki:
I see that it's moving which is a good news. @Tapchicoma any idea when it can be merged / released?

Tapchicoma:
@pbielicki haven't had time before, but has a plan to address review comments till Friday.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/LoadReportersTask.kt
Tapchicoma: Fixed via https://github.com/JLLeitschuh/ktlint-gradle/pull/424/commits/f8844a8faa4198e57ed24198cf947b5de67f03bf

hiljusti:
If time is the bottleneck for these last bits, let me know. I could carve out some time next week

Tapchicoma:
@JLLeitschuh 

> Add archunt tests to verify that ktlint classes aren't called where they shouldn't be.

Please check https://github.com/JLLeitschuh/ktlint-gradle/pull/424/commits/c899d633795732bc33c0a3482ae58c4aa9f93f27 commit.

> Do the finalizeBy thing suggested by Paul.

From my tests if worker in linting task fails the build, finalizer task is not called (from documentation I would expected otherwise). I've asked related question in Gradle slack https://gradle-community.slack.com/archives/CA745PZHN/p1611928971081900, but nobody answered if it is expected behaviour :man_shrugging: 

TLDR: I would propose to leave as it is currently.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/worker/SerializableReporterProviderTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.ObjectInputStream
+import java.io.ObjectOutputStream
+import java.io.PrintStream
+
+internal class SerializableReporterProviderTest {
+    @TempDir
+    lateinit var temporaryFolder: File
+
+    @Test
+    internal fun `Should correctly serialize and deserialize ReporterProvider`() {
```
JLLeitschuh, line 20: Don't feel obligated to mark test classes/methods as `internal`. The default access modifiers are completely fine.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt
```diff
@@ -37,24 +42,93 @@ private fun resolveGroup(ktlintVersion: String) = when {
...
+                        // project is from the current build.
+                        target.dependencies.add(
+                            configName,
+                            target.dependencies.project(mapOf("path" to id.projectPath))
+                        )
+                    }
+                }
+            }
+        }
+    }
```
JLLeitschuh, line 133: Does this have some kinds of test coverage verifying that it works as expected?

JLLeitschuh (commented):
I think just one final question and we can merge this.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt
```diff
@@ -37,24 +42,93 @@ private fun resolveGroup(ktlintVersion: String) = when {
...
+                        // We should avoid doing this for composite builds, so we check if the selected
+                        // project is from the current build.
+                        target.dependencies.add(
+                            configName,
+                            target.dependencies.project(mapOf("path" to id.projectPath))
+                        )
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Tapchicoma: Yes, check [this test](https://github.com/Tapchicoma/ktlint-gradle/blob/c899d633795732bc33c0a3482ae58c4aa9f93f27/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt#L462-#L477)

By running on lowest supported and current one Gradle versions - both branches are been verified.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt
Tapchicoma: Duplicated question, answered in the first one.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt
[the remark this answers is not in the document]
JLLeitschuh: Sounds good!

JLLeitschuh (approved):
LGTM!

Tapchicoma:
For those who is interested - I will try to push new release tomorrow.

PR #425, opened 2021-01-04
Kotlin DSL example sections updated

artour-bakiev:
Now the Kotlins DSL examples show the proper way to configure `ktlint` tasks


Tapchicoma (approved):
Thank you for the fix :+1:

PR #431, opened 2021-02-01
Requirement for GIT Pre-Commit Hooks Added

StuStirling:
Made it more obvious that both plugins are needed to have the extra gradle tasks related to Idea and GIT


Tapchicoma (approved):
Thank you for clarifying it :+1:

PR #435, opened 2021-02-08
Ensure Kotlin script files from subprojects are not checked by parent project.

### LINKED ISSUE #434: runKtlintCheckOverKotlinScripts & runKtlintFormatOverKotlinScripts process too many files
Currently,  when you utilize the `runKtlintCheckOverKotlinScripts` and `runKtlintFormatOverKotlinScripts` tasks, they, by default configure a file tree with the wildcard `**/*.kts`.

Given the following file structure, where `ktlint-gradle` is configured on the root project and all sub-projects.

```
parent-build.gradle.kts
subproject
  - child-build.gradle.kts
```

In this situation, `child-build.gradle.kts` is linted/formatted twice. Once by the root project, and once by the child project.

This has further implications.

In Gradle 7, some checks were introduced to check if two tasks have overlapping inputs/outputs that are undeclared.

If we specify that the `ktlint-gradle` input is `**/*.kts`, we suddenly have a `FileTree` that is rooted in the root, and all subdirectories, including the `build` directory, become inputs to our task. Thus our plugin will have input overlap with all tasks.

---

I believe that the fix here is to change the input wildcard to simply be `*.kts`. This forces users to manually specify sub-directories that contain script files, for example if scripts are in the `gradle` directory. But I think this change brings us closer to being correct.

CC: @Tapchicoma 

Credit to @wolfs for helping explain this to me.

Tapchicoma:
Should fix #434

### DIFF
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinScriptTest.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinScriptTest.kt
@@ -1,0 +1,54 @@
+package org.jlleitschuh.gradle.ktlint
+
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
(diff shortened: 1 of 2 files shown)


Tapchicoma:
Not an issue.

PR #468, opened 2021-04-12
Remove all usages of 'jcenter()' repository.

### LINKED ISSUE #460: JCenter Dependency Removal before JCenter Service Brown-Out Periods
JCenter has announced it will no longer be available in the near future. Their timeline specifies that they will have service brown-out  outages between April 12th to April 26th, 2021. During this period, there will be various outages that can cause various build failures.

I've been testing to find dependencies and plugins that still use JCenter by re-routing JCenter connections to local host cause a resolution failure, and came across this dependency running into an error. (My repo uses MavenCentral)
Screen Shot 2021-03-26 at 1 42 42 PM

I've noticed in this repository, a recent commit has been made to remove JCenter: https://github.com/JLLeitschuh/ktlint-gradle/commit/09fcfa7d91cdbe1ed853a78ee6ac59fcdf318774

I was wondering if there's a timeline for when this change will be released? Or if there's a solution that could be suggested for the time being. I'm currently using an older version of this plugin.

Thanks!

Tapchicoma:
Fixes #460

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,6 @@ allprojects {
     repositories {
         google()
         mavenCentral()
-        jcenter() // required for old html reporter versions
     }
 }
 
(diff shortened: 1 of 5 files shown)


JLLeitschuh (approved):
LGTM!

PR #469, opened 2021-04-17
Reset KtLint internal cache on .editorconfig files change

### LINKED ISSUE #456: Changes in .editorconfig file are ignored on subsequent reruns
After updating to ktlint 0.41.0 I get:
```
A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.KtLintWorkAction
   java.lang.RuntimeException: Unexpected imports layout: "*,java.**,javax.**,kotlin.**,^"
```

I know about the changes made regarding https://github.com/pinterest/ktlint/pull/1018 and tried different configurations in my .editorconfig. But non of these seems to work:

> kotlin_imports_layout=idea
> ij_kotlin_imports_layout=*
> ij_kotlin_imports_layout=*,java.**,javax.**,kotlin.**,^

Not sure if this is an incompatibility with the plugin oder a general ktlint issue.

Plugin Version: 10.0.0
Ktlint Version: 0.41.0
Gradle Version: 6.8.3
Android Studio: 4.1.2

Tapchicoma:
Reset internal KtLint caches on any `.editorconfig` files change, so KtLint will repopulate caches and takes into account new content.

Additionally:
- slightly optimize `.editorconfig` files locate logic
- re-run KtLint on all sources on any `.editorconfig` files change, as change may affect rules behaviour

Fixes #456

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
@@ -49,10 +49,14 @@ abstract class BaseKtLintCheckTask @Inject constructor(
     @Suppress("unused")
     @get:PathSensitive(PathSensitivity.RELATIVE)
     @get:InputFiles
-    internal val editorConfigFiles: FileCollection by lazy(LazyThreadSafetyMode.NONE) {
-        // Gradle will lazy evaluate this task input only on task execution
-        getEditorConfigFiles(project, additionalEditorconfigFile)
-    }
+    internal val editorConfigFiles: FileCollection = objectFactory.fileCollection().from(
+        {
+            getEditorConfigFiles(
+                projectLayout.projectDirectory.asFile.toPath(),
+                additionalEditorconfigFile
+            )
+        }
+    )
 
     @get:Input
     internal abstract val ktLintVersion: Property<String>
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
@@ -5,6 +5,7 @@ import org.gradle.api.GradleException
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileTree
+import org.gradle.api.file.FileType
 import org.gradle.api.file.ProjectLayout
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.model.ObjectFactory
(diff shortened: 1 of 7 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -141,9 +163,28 @@ abstract class BaseKtLintCheckTask @Inject constructor(
             params.formatSource.set(formatSources)
             params.discoveredErrorsFile.set(discoveredErrors)
             params.ktLintVersion.set(ktLintVersion)
+            params.editorconfigFilesWereChanged.set(editorConfigUpdated)
         }
     }
 
+    private fun wereEditorConfigFilesUpdated(
```
JLLeitschuh, line 170:
```suggestion
    private fun wasEditorConfigFilesUpdated(
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -116,11 +121,28 @@ abstract class BaseKtLintCheckTask @Inject constructor(
         )
 
     protected fun runLint(
-        filesToCheck: Set<File>,
+        inputChanges: InputChanges?,
         formatSources: Boolean,
     ) {
         checkDisabledRulesSupportedKtLintVersion()
 
+        val editorConfigUpdated = wereEditorConfigFilesUpdated(inputChanges)
```
JLLeitschuh, line 129:
```suggestion
        val editorConfigUpdated = wasEditorConfigFilesUpdated(inputChanges)
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
```diff
@@ -75,6 +81,13 @@ abstract class KtLintWorkAction : WorkAction<KtLintWorkAction.KtLintWorkParamete
             )
     }
 
+    private fun resetEditorconfigCache() {
+        if (parameters.editorconfigFilesWereChanged.get()) {
+            logger.info("Resetting KtLint caches")
+            KtLint.trimMemory()
```
JLLeitschuh, line 87: Can you add some documentation on why this is the method to call? `trimMemory` seems like a terrible method name for something like this, but maybe I'm misunderstanding something here.

JLLeitschuh (approved):
A few suggestions of name changes and one bit of documentation inline that would be appreciated.

Nice work on this. Thanks!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
Tapchicoma: Basically this method call means to reduce KtLint internal memory caches that is also cleaning loaded `.editorconfig` cache. Will add a comment.

PR #474, opened 2021-04-28
Add git hook version

KotlinIsland:
When git hook is updated users won't know. This change introduces a version number into the hook, and if that version number doesn't match the expected value then the gradle task will fail.


KotlinIsland:
I'm not sure about the best approach for the version of the git hook, I would think matching the version of ktlint would be a bad idea as it would force users to update the hook when there are no changes.

Currently this requires manually updating the version when changes are made to GitHook.kt

Perhaps the best way to solve this would be to have a build step that checks for changes to GitHook.kt or generates the script and checks diff. If there are changes then increment the `hookVersion`.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -89,6 +90,10 @@ abstract class BaseKtLintCheckTask @Inject constructor(
 
     init {
         if (project.hasProperty(FILTER_INCLUDE_PROPERTY_NAME)) {
+            // if FILTER_INCLUDE_PROPERTY_NAME exists then we are invoked from a git hook, check hook version
+            if (project.findProperty(hookVersion) != hookVersion) {
+                throw GradleException("Your ktlint git hook is outdated, please update by running the addKtlint*GitPreCommitHook Gradle task.")
+            }
```
JLLeitschuh, line 96: I'd be inclined to make this a warning instead of an error. Or, if we do want to make this an error, you should be able to disable it to make it a warning instead.

@Tapchicoma thoughts?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -66,28 +79,29 @@ internal fun generateGitHook(
...
+    
     echo "Completed ktlint run."
+    
     ${postCheck(shouldUpdateCommit)}
     
     git stash pop
     
     echo "Completed ktlint hook."
-
+
```
JLLeitschuh, line 104: A bunch of indentation was added here. Can you trim this out?

KotlinIsland: Some lines already have indentation, do you want me to remove it from all lines or leave as is?

JLLeitschuh: Can you trim the new indents?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
KotlinIsland: I've changed this to be a warning.

KotlinIsland: If it's a warning then the user won't see it because `--quiet`.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/GitHookTasksTest.kt
```diff
@@ -271,6 +271,23 @@ class GitHookTasksTest : AbstractPluginTest() {
...
+    @Test
+    internal fun `Git hook should send the hook version to gradle`() {
+        projectRoot.setupGradleProject()
+        val gitDir = projectRoot.initGit()
+
+        build(":$INSTALL_GIT_HOOK_FORMAT_TASK").run {
+            assertThat(task(":$INSTALL_GIT_HOOK_FORMAT_TASK")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+            assertThat(gitDir.preCommitGitHook().readText()).contains("""-phookVersion=$hookVersion""")
+        }
+    }
```
JLLeitschuh, line 283: Could you update this test to use the `@CommonTest` format above?

JLLeitschuh (changes requested):

PR #475, opened 2021-04-29
Add KtLint baseline support

### LINKED ISSUE #414: Add support for ktlint baseline
Ktlint https://github.com/pinterest/ktlint
> to add a baseline to check against prepend following args: "--baseline=ktlint-baseline.xml"

Tapchicoma:
This PR adds initial KtLint baseline support with following limitations:
- Format tasks ignore baseline, see [KtLint 1072 issue](https://github.com/pinterest/ktlint/issues/1072) for more details
- One baseline file is generated per one Gradle project (module). Merging baseline is not possible as KtLint actually uses `Reporter` interface to generate baseline file. Theoretically it will be possible to introduce special generate task in the root project that will trigger all subprojects, collects all errors and then write one baseline file, but let us see how big demand for it will be.

Fixes #414

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
@@ -108,13 +108,13 @@ internal constructor(
     /**
      * Baseline file location.
      *
-     * Default location is `<projectDir>/ktlintBaseline.xml`.
+     * Default location is `<projectDir>/config/ktlint/baseline.xml`.
      *
      * @since KtLint `0.41.0`
      */
     val baseline: RegularFileProperty = objectFactory.fileProperty()
         .convention(
-            projectLayout.projectDirectory.file("ktlintBaseline.xml")
+            projectLayout.projectDirectory.dir("config").dir("ktlint").file("baseline.xml")
         )
 
     private val kscriptExtension = KScriptExtension(kotlinScriptAdditionalPathApplier)
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
@@ -4,6 +4,7 @@ import groovy.lang.Closure
 import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.file.ConfigurableFileTree
+import org.gradle.api.file.ProjectLayout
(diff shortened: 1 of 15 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -103,6 +105,18 @@ internal constructor(
...
+    /**
+     * Baseline file location.
+     *
+     * Default location is `<projectDir>/ktlintBaseline.xml`.
+     *
+     * @since KtLint `0.41.0`
+     */
+    val baseline: RegularFileProperty = objectFactory.fileProperty()
+        .convention(
+            projectLayout.projectDirectory.file("ktlintBaseline.xml")
```
JLLeitschuh, line 117: Would it be better to make the convention having this inside of a directory instead of in the project directory?

JLLeitschuh (changes requested):

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
Tapchicoma: which directory do you have in mind? Something like `$projectDir/ktlint/ktlintBaseline.xml`?

Tapchicoma: Have changed it to `$projectDir/config/ktlint/baseline.xml`

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/UnsupportedGradleTest.kt
```diff
@@ -14,6 +16,7 @@ class UnsupportedGradleTest : AbstractPluginTest() {
             .withGradleVersion("5.6.4")
 
     @Test
+    @DisabledOnOs(OS.WINDOWS)
```
JLLeitschuh, line 19: What's up here?

JLLeitschuh (approved):
LGTM!

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/UnsupportedGradleTest.kt
Tapchicoma: Gradle test daemon disappearing on Github Windows runner. I've tried to reproduce it on another windows installation, but test there is passing successfully.

And actually it is ok to run this test only on Linux runner.

PR #482, opened 2021-05-12
Updating references to MPP samples

neugartf:
Just fixing 404s.


Tapchicoma:
Thank you for fixing it :+1:

PR #502, opened 2021-06-15
Resolve pre-commit hook causing conflicts

### LINKED ISSUE #497: Precommit hook causes conflicts and doesn't commit changes
With 10.1.0, and the latest precommit hook from 10.1.0.

1. Make a badly-formatted change to a file.
2. Stage the change.
3. Commit.
4. Observe that the change has not been committed, and the file is now showing a conflict.

CleanShot 2021-06-07 at 11 34 40@2x

KotlinIsland:
resolves #497

The logic was using `git stash push --keep-index` which keeps the index but also adds it to the stash. I've change the functionality to use a .diff file with the changes `git diff > unstaged.diff` and reapply it with `git apply --ignore-whitespace unstaged.diff`, in my opinion using `-3` would be better but then git applies it to the index, not the working copy.

There isn't a current version in the changes so I bumped it to 10.1.1

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
@@ -77,14 +77,22 @@ internal fun generateGitHook(
     echo "Running ktlint over these files:"
     echo "${'$'}CHANGED_FILES"
     
-    git stash push --keep-index
+    diff=.git/unstaged-ktlint-git-hook.diff
+    git diff --color=never > ${'$'}diff
+    if [ -s ${'$'}diff ]; then
+      git apply -R ${'$'}diff
+    fi
     
     ${generateGradleCommand(taskName, gradleRootDirPrefix)}
 
     echo "Completed ktlint run."
     ${postCheck(shouldUpdateCommit)}
     
-    git stash pop
+    if [ -s ${'$'}diff ]; then
+      git apply --ignore-whitespace ${'$'}diff
+    fi
+    rm ${'$'}diff
+    unset diff
     
     echo "Completed ktlint hook."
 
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -77,14 +77,20 @@ internal fun generateGitHook(
     echo "Running ktlint over these files:"
     echo "${'$'}CHANGED_FILES"
     
-    git stash push --keep-index
+    git diff > unstaged.diff
```
Tapchicoma, line 80: how about moving `unstaged.diff` into `.git/` directory and name it something like `unstaged-ktlint-git-hook.diff` to make it have a little more unique name and reduce possible other hooks name conflict?

Tapchicoma: Probably would be nice additionally introduce some variable to hold diff file name.

Tapchicoma: And better to add `--color=never` diff option, so produced patch could always be consumed by `git apply` command

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/GitHookTasksTest.kt
```diff
@@ -157,8 +157,10 @@ class GitHookTasksTest : AbstractPluginTest() {
         build(":$INSTALL_GIT_HOOK_FORMAT_TASK").run {
             assertThat(task(":$INSTALL_GIT_HOOK_FORMAT_TASK")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
             val hookText = gitDir.preCommitGitHook().readText()
-            assertThat(hookText).contains("git stash push")
-            assertThat(hookText).contains("git stash pop")
+            assertThat(hookText).contains("git diff > unstaged.diff")
```
Tapchicoma, line 160: I am really need to come up with git hook test setup that actually runs installed githook on some test project :thinking:

Tapchicoma (commented):
Could you rebase your PR against `master` and put your changes into '[Unreleased]' changelog section?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/GitHookTasksTest.kt
KotlinIsland: #503

KotlinIsland:
Only thing i don't like is if the patch fails to apply after format, `ignore-whitespace` will help but is that good enough?

Tapchicoma (approved):

PR #513, opened 2021-07-16
Gradle Toolchain support

rpalcolea:
Hi folks,

Wondering if this change makes sense to support Toolchains


Tapchicoma:
I highly doubt KtLint depends on JDK version it is running on to add toolchain support:
- KtLint uses Kotlin compiler to parse AST, but Kotlin compiler itself could run on any supported JDK version and produce same result (as long as optional `--jdk-home` option has the same value)
- AFAIK KtLint rules does not have anything JDK version specific

JLLeitschuh:
Closing  for now. If there's a strong need for this feature, then we're happy to reconsider adding this. Anyone who really needs this, please don't hesitate to provide insights into a larger use case for this feature

PR #515, opened 2021-07-18
Prevent creation of empty directories on running format task.

### LINKED ISSUE #423: ktlintFormat creates directories under src
When I run `ktlintFormat` there are the following directories, which are created under `src` directory:
- src/main/java
- src/main/kotlin
- src/test/java
- src/test/kotlin

It is pretty annoying seeing java directory all the time and have no chance to get rid of it when you write a 100% Kotlin project. Is there a way to ask ktlint not to create these directories?

Tapchicoma:
Now base lint task does not extend SourceTask, but DefaultTask + PatternFilterable. This allows for format task to filter empty directories in output, so Gradle will not create empty SourceSet directories.

Fixes #423


Tapchicoma:
Same changes in #520 PR

PR #516, opened 2021-07-19
Added support for sarif reporter

abs-paradigm:
Added support for the ktlint sarif reporter: https://github.com/pinterest/ktlint/pull/1113

It only available in the 0.42.0-SNAPSHOT ktlin version at moment.


Tapchicoma (approved):
Thank you for adding such support :+1:

PR #520, opened 2021-08-08
Rethink format task approach

### LINKED ISSUE #306: Rethink format task approach
Currently format tasks works the same as common lint tasks, except that they may modify inputs. This leads to the several problems with this tasks caching: #302 or #194 

Possible solutions to this problem:
- actually run lint tasks first, don't fail the tasks, then run format only on files with files violations. Format task is never cacheable, but uses lint tasks reports as input.
- something else?

@JLLeitschuh maybe something else comes to your mind how to solve it as well?

### LINKED ISSUE #423: ktlintFormat creates directories under src

Tapchicoma:
`BaseKtlintCheckTask` now extends `DefaultTask` plus implements `PatternFilterable` interface copying approach from `SourceTask`.

To better handle incremental formatting, plus handle the case of restored to pre-formatted state sources, Format task tracks pre-formatted files snapshot: file + content hash. In case sources has the files with same content hash as file in snapshot from previous task execution - format tasks fails up-to-date check and run format again on such files.

Fixes #423
Fixes #306

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   - Update Shadow plugin to `7.0.0` version
   - Update Kotlin to `1.5.21` version
   - Set default KtLint version to `0.42.1`
+  - Rethink format task approach ([issue: #306](https://github.com/JLLeitschuh/ktlint-gradle/issues/306))
 
 ### Fixed
   - Pre-commit hook causing conflicts ([issue: #443](https://github.com/JLLeitschuh/ktlint-gradle/issues/443)) ([#502](https://github.com/JLLeitschuh/ktlint-gradle/pull/502))
-  - `ktlintFormat` create empty directories in `src/` dir ([issue: $423](https://github.com/JLLeitschuh/ktlint-gradle/issues/423))
+  - `ktlintFormat` create empty directories in `src/` dir ([issue: #423](https://github.com/JLLeitschuh/ktlint-gradle/issues/423))
 
 ### Removed
   - ?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
   - Pre-commit hook causing conflicts ([issue: #443](https://github.com/JLLeitschuh/ktlint-gradle/issues/443)) ([#502](https://github.com/JLLeitschuh/ktlint-gradle/pull/502))
+  - `ktlintFormat` create empty directories in `src/` dir ([issue: $423](https://github.com/JLLeitschuh/ktlint-gradle/issues/423))
 
 ### Removed
   - ?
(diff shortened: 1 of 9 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -87,6 +92,9 @@ abstract class BaseKtLintCheckTask @Inject constructor(
         convention("256m")
     }
 
+    private var sourceFiles: ConfigurableFileCollection = objectFactory.fileCollection()
+    private val patternSet: PatternFilterable = PatternSet()
```
JLLeitschuh, line 96: Can an instance be injected? I don't think the constructor will be change, but if it is, this way we're forwards compatible.

CHANGELOG.md
```diff
@@ -13,9 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   - Update Gradle to `7.1.1` version
   - Update Shadow plugin to `7.0.0` version
   - Update Kotlin to `1.5.21` version
+  - Rethink format task approach ([issue: $306](https://github.com/JLLeitschuh/ktlint-gradle/issues/306))
 
 ### Fixed
   - Pre-commit hook causing conflicts ([issue: #443](https://github.com/JLLeitschuh/ktlint-gradle/issues/443)) ([#502](https://github.com/JLLeitschuh/ktlint-gradle/pull/502))
+  - `ktlintFormat` create empty directories in `src/` dir ([issue: $423](https://github.com/JLLeitschuh/ktlint-gradle/issues/423))
```
JLLeitschuh, line 20: Is this syntax correct? Shouldn't it be this?
```suggestion
  - `ktlintFormat` create empty directories in `src/` dir ([issue: #423](https://github.com/JLLeitschuh/ktlint-gradle/issues/423))
```
Same question about the line above

JLLeitschuh (changes requested):
Looks good to me other than the issue in the changelog

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
Tapchicoma: Unfortunately `PatternFilterable` is not injectable by Gradle, so I changed the code to pass `PatternSet` as a constructor parameter.

CHANGELOG.md
Tapchicoma: Fixed the typo!

Tapchicoma:
@JLLeitschuh could you check updated PR?

JLLeitschuh (approved):
LGTM!

PR #534, opened 2021-09-15
Bump github actions

Goooler:


JLLeitschuh (approved):
Looks good to me. @Tapchicoma feel free to merge if you're good with this change. Thanks @Goooler!

.github/workflows/build-and-check.yml
```diff
@@ -15,13 +15,14 @@ jobs:
...
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: set up JDK 1.8
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.8
+          java-version: 8
```
nuhkoca, line 22: Can't `JDK 11` be used instead?

Goooler: Sure, we can.

JLLeitschuh: I'd prefer if we continued to run with JDK 1.8 as a minimum for now and try adding JDK 11 as a matrix version. I want to make sure we don't break JDK 1.8 support, but also testing for JDK 11 support would be good.

JLLeitschuh: Don't feel obligated to add JDK 11 testing support in this PR. If you just want to leave it at version 8, that's fine.

nuhkoca: @JLLeitschuh sure thing! We can definitely keep it out of scope of this PR but I'd like to take it up to support `JDK 11`, too. WDYT?

Tapchicoma: Personally, I would prefer leaving here JDK 1.8 same as Gradle minimal supported JDK version.

PR #558, opened 2021-12-27
Fix src/main/kotlin not being recognised by default

### LINKED ISSUE #524: src/main/kotlin not working out of the box
As of Android Studio Arctic Fox and Android Gradle Plugin 7.0, we do not need to specify the source directories for java:

https://stackoverflow.com/questions/41467453/renaming-java-directory-to-kotlin-in-android-studio

You should also support this out of the box for ktlint.

Goooler:
Closes #524.

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -120,3 +120,33 @@ jobs:
             ${{ runner.os }}-gradle-wrapper-
       - name: Check ktlint-plugin on samples
         run: ./gradlew ktlintCheck --no-daemon -x :sample:android-app:ktlintCheck :sample:kotlin-mpp-android:ktlintCheck
+
+  check_android_samples:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+    runs-on: ${{ matrix.os }}
+    needs: build
+    steps:
+      - uses: actions/checkout@v2
+      - name: set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: 'zulu'
+      - name: Restore Gradle caches
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-cache-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-cache-
+      - name: Restore Gradle wrapper
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/wrapper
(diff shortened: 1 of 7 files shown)


Goooler:
```
KtLintSupportedVersionsTest > Format should successfully finish on sources with style violations > org.jlleitschuh.gradle.ktlint.KtLintSupportedVersionsTest.Format should successfully finish on sources with style violations$ktlint_gradle(GradleVersion, String)[9] FAILED
    java.lang.IllegalStateException at KtLintSupportedVersionsTest.kt:80
        Caused by: org.gradle.tooling.GradleConnectionException at KtLintSupportedVersionsTest.kt:80
            Caused by: org.gradle.launcher.daemon.client.DaemonDisappearedException at DaemonClient.java:249
```

Like an unrelated failure.

.github/workflows/build-and-check.yml
```diff
@@ -67,10 +67,10 @@ jobs:
     needs: build
     steps:
       - uses: actions/checkout@v2
-      - name: set up JDK 1.8
+      - name: set up JDK 11
         uses: actions/setup-java@v2
         with:
-          java-version: 8
+          java-version: 11
           distribution: 'zulu'
```
JLLeitschuh, line 74: We'd like to maintain backwards compatibility with JDK 8, so building with 8 is what we want still.

However, if you'd like to add support for matrix testing with 8, 11, and 17, that would be cool. If you do do that, please make sure that publishing still happens with JDK 8

Goooler: But the min JDK required by AGP is 11 now.

```
* What went wrong:
An exception occurred applying plugin request [id: 'com.android.application', artifact: 'com.android.tools.build:gradle:7.0.4']
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 11 to run. You are currently using Java 1.8.
     You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

JLLeitschuh: Can you reconfigure the build so that the android tests only execute when executed with the required java version? I'd still like to test the plugin against jdk 1.8 for non-android purposes. Also, does this change mean that we'll drop support for older versions of the Android plugin or are these changes backwards compatible?

@Tapchicoma you understand this much better than I do, can you weigh in here?

Goooler: 😅 No idea.

Goooler: I've add `check_android_bashisms` & `check_android_samples` jobs to check android modules, CC @JLLeitschuh

Goooler:
This should be fixed by #733.

Issue #562, opened 2022-01-05
Pre Commit Hook Error


PaulWoitaschek:
When running the precommit hook it always fails with the message:

```
.git/hooks/pre-commit: line 22: =0: command not found
```

This is probably because of this line:

```
$gradleCommandExitCode=$?
```
Imo the trailing $ should be removed.

This also has the very bad sideffect that the old stash is not applied and the code that wasn't staged is lost.

Tested on macos, Ktlint 10.2.1

jmbarroso:
Found the same problem. Edited manually the file and it worked as expected

PauloHC:
I am getting the same error. I didn't find this line after applying the plugin to the project.
Is there any workaround to make the pre-commit hook work fine?

jmbarroso:
@PauloHC, after installing the pre-commit, it should be available in `.git/hooks/pre-commit`. Line 22 should look like `$gradleCommandExitCode=$?` . The workaround  is to  change it to `gradleCommandExitCode=$?`

PauloHC:
I found the line here. Now the hook is working. Thank you very much @jmbarroso !

G00fY2:
This behaviour was changed in #551. We should try to fix it not just deleting this line.

PR #564, opened 2022-01-08
Fix pre-commit hook command not found error

rbleuse:
Fixes https://github.com/JLLeitschuh/ktlint-gradle/issues/562


G00fY2:
@rbleuse I noticed your PR too late. Please check out my changes from #568. You don't need Regex to improve the tests. I also suggest to switch to switch to underscore_case for consistency (e.g. CHANGED_FILES).

rbleuse:
Hi @G00fY2 thanks for the suggestion !

G00fY2:
Not sure why the tests on windows are not passing. Looks like they are somewhat flaky. I tried to reproduce this on my windows machine but actually for me all tests are failing (maybe because I use JDK 17).

JLLeitschuh:
Here's the exception, I don't think this was caused by you. I've re-kicked the jobs:

https://scans.gradle.com/s/zjxfgwkm3i2fg/tests/:test/org.jlleitschuh.gradle.ktlint.KtLintSupportedVersionsTest/Format%20should%20successfully%20finish%20on%20sources%20with%20style%20violations$ktlint_gradle(GradleVersion,%20String)%5B12%5D?anchor=e30&expanded-stacktrace=WyIwLTEtMiIsIjAtMSIsIjAiXQ&top-execution=1

rbleuse:
It seems it failed again :

```
2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':samples:android-app:loadKtlintReporters'.
> Could not resolve all files for configuration ':samples:android-app:ktlint'.
   > Could not resolve com.pinterest:ktlint:0.42.1.
     Required by:
         project :samples:android-app
      > Could not resolve com.pinterest:ktlint:0.42.1.
         > Could not get resource 'https://repo.maven.apache.org/maven2/com/pinterest/ktlint/0.42.1/ktlint-0.42.1.pom'.
            > Could not GET 'https://repo.maven.apache.org/maven2/com/pinterest/ktlint/0.42.1/ktlint-0.42.1.pom'.
               > repo.maven.apache.org
```

crisguitar:
Any reasons this is not merged yet?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -54,13 +54,12 @@ private fun postCheck(
             git add ${'$'}file
         fi
     done
-    """.trimIndent()
```
JLLeitschuh, line 57: Doesn't some kind of trimming need to happen here?

JLLeitschuh:
> Any reasons this is not merged yet?

The tests were failing. I was waiting for a green run

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
rbleuse: @JLLeitschuh No because if we keep it, `formatUsesGradleExitCode` fails to find `gradle_command_exit_code=$?` as there is indentation before :

<details><summary>With trimming (original)</summary>

```
Expecting:
 <["#!/bin/sh",
 "######## KTLINT-GRADLE HOOK START ########",
 "",
 "    CHANGED_FILES="$(git --no-pager diff --name-status --no-color --cached | awk '$1 != "D" && $NF ~ /\.kts?$/ { print $NF }')"",
 "",
 "    if [ -z "$CHANGED_FILES" ]; then",
 "        echo "No Kotlin staged files."",
 "        exit 0",
 "    fi;",
 "",
 "    echo "Running ktlint over these files:"",
 "    echo "$CHANGED_FILES"",
 "",
 "    diff=.git/unstaged-ktlint-git-hook.diff",
 "    git diff --color=never > $diff",
 "    if [ -s $diff ]; then",
 "      git apply -R $diff",
 "    fi",
 "",
 "    ./gradlew --quiet ktlintFormat -PinternalKtlintGitFilter="$CHANGED_FILES"",
 "    gradle_command_exit_code=$?",
 "",
 "    echo "Completed ktlint run."",
 "    echo "$CHANGED_FILES" | while read -r file; do",
 "    if [ -f $file ]; then",
 "        git add $file",
 "    fi",
 "done",
 "",
 "    if [ -s $diff ]; then",
 "      git apply --ignore-whitespace $diff",
 "    fi",
 "    rm $diff",
 "    unset diff",
 "",
 "    echo "Completed ktlint hook."",
 "    exit $gradle_command_exit_code",
 "######## KTLINT-GRADLE HOOK END ########"]>
to contain:
 <["gradle_command_exit_code=$?"]>
but could not find:
 <["gradle_command_exit_code=$?"]>
```
</details>

<details><summary>Without trimming</summary>

```
"######## KTLINT-GRADLE HOOK START ########",
"",
"CHANGED_FILES="$(git --no-pager diff --name-status --no-color --cached | awk '$1 != "D" && $NF ~ /\.kts?$/ { print $NF }')"",
"",
"if [ -z "$CHANGED_FILES" ]; then",
"    echo "No Kotlin staged files."",
"    exit 0",
"fi;",
"",
"echo "Running ktlint over these files:"",
"echo "$CHANGED_FILES"",
"",
"diff=.git/unstaged-ktlint-git-hook.diff",
"git diff --color=never > $diff",
"if [ -s $diff ]; then",
"  git apply -R $diff",
"fi",
"",
"./gradlew --quiet ktlintFormat -PinternalKtlintGitFilter="$CHANGED_FILES"",
"gradle_command_exit_code=$?",
"",
"echo "Completed ktlint run."",
"",
"echo "$CHANGED_FILES" | while read -r file; do",
"    if [ -f $file ]; then",
"        git add $file",
"    fi",
"done",
"",
"",
"if [ -s $diff ]; then",
"  git apply --ignore-whitespace $diff",
"fi",
"rm $diff",
"unset diff",
"",
"echo "Completed ktlint hook."",
"exit $gradle_command_exit_code",
"######## KTLINT-GRADLE HOOK END ########"]>
```
</details>

But if you tell me that the first indentation with `trimIndent` (while it's supposed to remove indentation) is expected, then I'll add it back and look for `exit $gradle_command_exit_code` with indentation in the test instead.

craigberry1:
Hey @JLLeitschuh will this get merged soon? We just updated to `10.2.1` to fix https://github.com/JLLeitschuh/ktlint-gradle/issues/539 but now I'm running into this... thanks!

JLLeitschuh:
Happy to merge this. @rbleuse can you rebase your changes and I'll merge! 😄

rbleuse:
@JLLeitschuh done :)

JLLeitschuh:
#I've been returning these tests multiple times and every time they fail with an OOM error. I have no idea why, but master doesn't fail. This failure is consistently reproducible with your changes and I have no idea why.

rbleuse:
@JLLeitschuh When you twice merged into my fix, the two builds were successful :
2022-03-13_12h55_13

I'd can confirm that this OOM is not because of this fix.
I reverted my fix except the changelog, and it still fails with an OOM (you can check last commit build result). There's definitely an issue with the windows build as the ubuntu builds correctly.

My guess is, parallel tests are enabled, and somehow there's a memory issue with parallel tests which use too much memory

Also one note, tests always fail into OOM on the `KtLintSupportedVersionsTest` class. Probably something wrong in this test class or in the code that this class is testing

Reference failed builds :
https://github.com/JLLeitschuh/ktlint-gradle/runs/5527597501?check_suite_focus=true
https://github.com/JLLeitschuh/ktlint-gradle/runs/5527663939?check_suite_focus=true
https://github.com/JLLeitschuh/ktlint-gradle/runs/5514012872?check_suite_focus=true
https://github.com/JLLeitschuh/ktlint-gradle/runs/5527730803?check_suite_focus=true

G00fY2:
I can aggree. I also run these tests on my local Windows machine. No such error occurred.

JLLeitschuh:
> I can confirm that this OOM is not because of this fix.

I 100% agree with you, but unfortunately, I can't merge this PR knowing that it will break `master`, it would prevent me from releasing. Unfortunately, I've already spent more time trying to figure out what the problem is here, and haven't had any luck.

I'm trying one last thing with this change: https://github.com/JLLeitschuh/ktlint-gradle/commit/bbc0c9423d0cdd35df080963ea9c24eb71f6d3e3

If we can't figure out the root cause of this OOM issue I can't merge this PR. Unfortunately, as I spelled out in this issue here https://github.com/JLLeitschuh/ktlint-gradle/issues/569 I'm not really available anymore to maintain this project beyond merging PRs from the community. I don't actively use ktlint-gradle anymore, and I'm now a full-time security researcher not working on any Kotlin or Gradle based projects actively at this time. I'm more than happy to merge any PRs that adequately fix this overall OOM error, or even PRs that help debug what's going wrong, but I don't think I have the time to continue to figure out what's wrong here. I'm really sorry as that's probably pretty disappointing to hear.

rbleuse:
Yes I don't have much time to investigate where this OOM comes from either, but as I mentioned it doesn't come from this PR :( 

As far as we can investigate from builds history is that the PR #543 has also an OOM when the PR has been merged to master
Here is build log : https://github.com/JLLeitschuh/ktlint-gradle/runs/4739934010?check_suite_focus=true

Anyway, your last commit didn't pass either, so what should we do ?

JLLeitschuh:
> Anyway, your last commit didn't pass either, so what should we do ?

I'm not certain. Maybe this is caused by a specific version of the Ktlint library on windows? If so, and someone can figure that out, a PR to disable that in our tests would be accepted.

I just can't merge a PR I know breaks the main branch as I won't be able to publish releases. I'm really sorry.

rbleuse:
> I just can't merge a PR I know breaks the main branch as I won't be able to publish releases. I'm really sorry.

With all due respect, did you notice that almost all your `Create prepare_release.yaml` commits directly on master also failed with this same insufficient memory error ?
As I mentioned before, this error doesn't come from this PR, as main is already broken.
Besides, my PR now doesn't have any extra code, only the changelog (as it's required to start testing with GA) so it's in pair identical to the main and it also fails.

What about increasing tests memory ? Same as mentioned in the readme
```
tasks.withType<org.jlleitschuh.gradle.ktlint.tasks.BaseKtLintCheckTask> {
    workerMaxHeapSize.set("512m")
}
```

Or disable parallel tests ?

JLLeitschuh:
> With all due respect, did you notice that almost all your Create prepare_release.yaml commits directly on master also failed with this same insufficient memory error ?


I actually didn't. So that 100% confirms this isn't your change breaking things. Okay, I'll merge it. Thanks for pointing out my mistake here. Sorry.

JLLeitschuh:
I've removed the windows tests from the critical path, so the other checks should theoretically now run

rbleuse:
> > With all due respect, did you notice that almost all your Create prepare_release.yaml commits directly on master also failed with this same insufficient memory error ?
> 
> I actually didn't. So that 100% confirms this isn't your change breaking things. Okay, I'll merge it. Thanks for pointing out my mistake here. Sorry.

No worries :)
I'll then commit back the fix later on today, for now this PR contains only the changelog

rbleuse:
It should be ok now

PR #568, opened 2022-01-24
Fix error in pre-commit hook bash syntax and improve tests

### LINKED ISSUE #562: Pre Commit Hook Error

G00fY2:
Fixes #562 

Instead of creating a new local variable `gradle_command_exit_code` the script instead tried to set to value to the variable `$gradle_command_exit_code`.
I fixes the bad syntax and improved the tests. The tests should have actually detected this, but because the pre-commit file was read as a String (`readText()`) the contains assert check returned true because `$gradle_command_exit_code` contains `gradle_command_exit_code`. I switched the logic to `readLines()` to check if the file actually contains the exact line. By that I fixed a indentation issue caused by the nested `trimIndent()` call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
   - ?
 
 ### Fixed
+  - Pre-commit hook syntax causing `command not found` [issue: #562](https://github.com/JLLeitschuh/ktlint-gradle/issues/562), [#568](https://github.com/JLLeitschuh/ktlint-gradle/pull/568)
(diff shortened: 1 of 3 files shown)


G00fY2:
Duplicate of #562

PR #571, opened 2022-01-27
Configure configurations as resolvable non-consumable

### LINKED ISSUE #523: Plugin breaks aar module
When ktlint plugin is applied to a module that just holds AAR artifact (via `allprojects` block), it appears to break that module's output - build complains about classes from that module missing.

Steps to reproduce: 

1. Download and assemble [KtlintAar.zip](https://github.com/JLLeitschuh/ktlint-gradle/files/7013474/KtlintAar.zip)
2. Note that build will fail
3. Go to root `build.gradle` and comment out `apply plugin` line for `ktlint`
4. Assemble again
5. Build will work

tbroyer:
Keeping the default of being consumable exposes them to
variant-aware resolution when the project is depended
upon from another project.

Also configure the configurations as non-visible for
similar reasons: those are specific to this project's
build and there's no reason to consume them from
another project.

Fix #523

### DIFF
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
@@ -508,6 +508,51 @@ class KtlintPluginTest : AbstractPluginTest() {
         }
     }
 
+    @DisplayName("Should not leak KtLint as a variant into consuming projects")
+    @CommonTest
+    fun noLeakIntoConsumingProjects(gradleVersion: GradleVersion) {
+        project(gradleVersion) {
+            val producerDir = projectPath.resolve("producer").also { it.mkdirs() }
+            val consumerDir = projectPath.resolve("consumer").also { it.mkdirs() }
+            settingsGradle.appendText(
+                """
+
+                include ":producer"
+                include ":consumer"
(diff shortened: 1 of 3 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -508,6 +508,51 @@ class KtlintPluginTest : AbstractPluginTest() {
         }
     }
 
+    @DisplayName("Should not leak KtLint as a variant into consuming projects")
```
JLLeitschuh, line 511:
```suggestion
    /**
     * See: https://github.com/JLLeitschuh/ktlint-gradle/issues/523#issuecomment-1022522032
     */
    @DisplayName("Should not leak KtLint as a variant into consuming projects")
```

JLLeitschuh (changes requested):
Excellent!! This is exactly the cohesive contribution I appreciate! Thank you for the awesome test that so simply demonstrates this issue. Can you just add the change I suggest and then I'm happy to approve/merge this

tbroyer:
I also added a changelog entry. The branch should be opened to maintainers though, so feel free to directly make changes there.

JLLeitschuh (approved):
LGTM!

PR #573, opened 2022-03-10
Add relative option similar to CLI's --relative

pschichtel:
We use this gradle plugin in GitLab-CI and currently it always produces absolute paths which leads to broken links in the code quality report.

This PR adds the option `relative` analog to ktlint's CLI's `--relative` option, which does the same thing: Instead of outputting absolute paths in the report, the reports are made relative to the root project (ktlint make it relative to the workdir).

Currently this is only applied to the actual report file output, not the console output to still have clickable file paths in IntelliJ.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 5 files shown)


pschichtel:
@JLLeitschuh just a friendly ping to make sure this PR is not accidentally overseen (I tend to miss/not receive notifications).

JLLeitschuh:
I actually did miss this, could you rebase this PR for me?

pschichtel:
@JLLeitschuh done

pschichtel:
@JLLeitschuh I see the windows check failed, but it doesn't look like that is related to this change. can you confirm?

Issue #589, opened 2022-06-19
The plugin does not work when using Ktlint 0.46.0


mars885:
The Ktlint [0.46.0](https://github.com/pinterest/ktlint/releases/tag/0.46.0) has been released recently with breaking API changes that make using the plugin impossible. One such change is that Ktlint.Params class has been renamed into the [ExperimentalParams](https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt#L61).

As a result of this, when I run the `./gradlew ktlintCheck --continue` command, I get the following:
```
> Task :app:runKtlintCheckOverKotlinScripts FAILED
> Task :app:runKtlintCheckOverMainSourceSet FAILED

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':app:runKtlintCheckOverKotlinScripts'.
> A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.KtLintWorkAction
   > com/pinterest/ktlint/core/KtLint$Params

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':app:runKtlintCheckOverMainSourceSet'.
> A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.KtLintWorkAction
   > com/pinterest/ktlint/core/KtLint$Params

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org

BUILD FAILED in 2s
10 actionable tasks: 3 executed, 7 up-to-date
```
The versions used are the following:
- ktlint-gradle: `10.3.0`
- ktlint: `0.46.0`

JLLeitschuh:
@scana could you do so in such a way that preserves backwards compatibility by using reflection to call the old APIs?

scana:
@JLLeitschuh unfortunately not, because they have also changed `ReporterProvider` signature to `ReporterProvider<T : Reporter>`

jzeferino:
with:

tlint-gradle: 10.3.0
ktlint: 0.46.1

still getting the following exception:

[...]

PR #592, opened 2022-06-21
Set KtLint version to 0.46.1

nuhkoca:


G00fY2:
I don't think only bumping the version actually works out. Did you see #589?

Maybe there is even some more effort required to support newer and older ktlint versions. Or we just allow ktlint 0.46.+. Not sure.

nuhkoca:
@G00fY2 yeah you are right. My bad, I'll just close this and start migrating to [kotlinter](https://github.com/jeremymailen/kotlinter-gradle)

PR #605, opened 2022-09-07
Improve execution in other then main worktree

knobo:
The hook does not run when in other worktrees since .git is not located there.


JLLeitschuh:
Are you able to add a unit test for this?

knobo:
Are there any other unit tests for the pre-comit hook?

knobo:
I have made tests similar to existing tests.

PR #625, opened 2023-02-07
fix disabled_rules warning in 0.48+

### LINKED ISSUE #622: kitlint 0.48.1 and "Property 'ktlint_disabled_rules' is deprecated" warning
Running `11.1.0` of the plugin, with this config block:

```
    ktlint {
        version = '0.48.1'
        android = true
    }
```

Generates hundreds of the following warnings (wrapped to two lines here, appears as one line in the output).

```
Property 'ktlint_disabled_rules' is deprecated: Update all your all '.editorconfig' files.
   See https://pinterest.github.io/ktlint/faq/#why-is-editorconfig-property-disabled_rules-deprecated-and-how-do-i-resolve-this for more information
```

Per that link, instead of a single `disabled_rules` line:

```
disabled_rules=rule-1,rule-2
```

the new approach is:

```
ktlint_<rulesetId>_rule-1=disabled
ktlint_<rulesetId>_rule-2=disabled
```

This confused the hell out of me because my project has no `.editorconfig` file. After digging through the plugin code I think the culprit is https://github.com/JLLeitschuh/ktlint-gradle/blob/main/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintInvocation.kt#L171, which inserts a "fake" editorconfig with a `disabled_rules` entry.

That will need to change.

This will probably be a breaking change for users, because if they have e.g.:

```
    disabledRules = [
        'import-ordering',
        'chain-wrapping',
        'no-blank-line-before-rbrace'
    ]
```

they will need to prepend the ruleset ID to each of those rule names.

wakingrufus:
when disabled_rules are empty, don't set them on the user data to avoid ktlint deprecated warnings. all disabled rules in 0.48+ should be set in .editorconfig directly according to the ktlint docs https://pinterest.github.io/ktlint/faq/#why-is-editorconfig-property-disabled_rules-deprecated-and-how-do-i-resolve-this

resolve #622
test to reproduce issue #622


JLLeitschuh (approved):
LGTM!

PR #634, opened 2023-02-22
Work around reflective access errors on JVM 16+

### LINKED ISSUE #629: Gradle 8 support
Gradle 8 is now released: https://docs.gradle.org/8.0/release-notes.html
let's support it

tadfisher:
Fixes #629

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
@@ -4,6 +4,7 @@ import groovy.lang.Closure
 import net.swiftzer.semver.SemVer
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
+import org.gradle.api.JavaVersion
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileTreeElement
@@ -254,6 +255,12 @@ abstract class BaseKtLintCheckTask @Inject constructor(
             spec.classpath.from(ktLintClasspath, ruleSetsClasspath)
             spec.forkOptions { options ->
                 options.maxHeapSize = workerMaxHeapSize.get()
+
+                // Work around ktlint triggering reflective access errors from the embedded Kotlin
+                // compiler. See https://youtrack.jetbrains.com/issue/KT-43704 for details.
+                if (JavaVersion.current() >= JavaVersion.VERSION_16) {
+                    options.jvmArgs("--add-opens", "java.base/java.lang=ALL-UNNAMED")
+                }
             }
         }
 
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/testAnnotations.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/testAnnotations.kt
@@ -12,10 +12,17 @@ import kotlin.streams.asStream
 
 object TestVersions {
     const val minSupportedGradleVersion = KtlintBasePlugin.LOWEST_SUPPORTED_GRADLE_VERSION
-    const val maxSupportedGradleVersion = "7.6"
+    const val maxSupportedGradleVersion = "8.0.1"
     val pluginVersion = File("VERSION_CURRENT.txt").readText().trim()
     const val minSupportedKotlinPluginVersion = "1.4.32"
-    const val maxSupportedKotlinPluginVersion = "1.5.31"
+
+    fun maxSupportedKotlinPluginVersion(
+        gradleVersion: GradleVersion = GradleVersion.version(maxSupportedGradleVersion)
+    ): String = when (gradleVersion) {
+        in (GradleVersion.version("6.8.3")..GradleVersion.version(maxSupportedGradleVersion)) -> "1.8.10"
+        in (GradleVersion.version("6.7.1")..GradleVersion.version("7.1.1")) -> "1.7.21"
+        else -> minSupportedKotlinPluginVersion
+    }
 }
 
 @Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ### Changed
 
 -   warn when additionalEditorconfigFile is used in 0.47+ [#637](https://github.com/JLLeitschuh/ktlint-gradle/pull/637)
+-   work around reflection error on Gradle 8/JDK 16+ [#634](https://github.com/JLLeitschuh/ktlint-gradle/pull/634)
 -   add ktlint version 0.48.2 to testing [#632](https://github.com/JLLeitschuh/ktlint-gradle/pull/632)
 -   update latest gradle version for testing to 7.6 [#632](https://github.com/JLLeitschuh/ktlint-gradle/pull/632)
 -   improve release process to update VERSION_LATEST_RELEASE automatically [#631](https://github.com/JLLeitschuh/ktlint-gradle/pull/631)
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KotlinJsPluginTests.kt
@@ -26,7 +26,7 @@ class KotlinJsPluginTests : AbstractPluginTest() {
             """
             
             kotlin {
-                js {
+                js(IR) {
                     nodejs()
                 }
             }
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintIdeaPluginTest.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintIdeaPluginTest.kt
@@ -24,9 +24,9 @@ class KtlintIdeaPluginTest : AbstractPluginTest() {
                             mavenLocal()
                             gradlePluginPortal()
                         }
-                        
+
                         plugins {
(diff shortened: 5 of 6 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -254,6 +255,12 @@ abstract class BaseKtLintCheckTask @Inject constructor(
             spec.classpath.from(ktLintClasspath, ruleSetsClasspath)
             spec.forkOptions { options ->
                 options.maxHeapSize = workerMaxHeapSize.get()
+
+                // Work around ktlint triggering reflective access errors from the embedded Kotlin
+                // compiler. See https://youtrack.jetbrains.com/issue/KT-43704 for details.
+                if (JavaVersion.current() >= JavaVersion.VERSION_16) {
```
wakingrufus, line 261: I am not that familiar with the  Worker API yet. are we sure this is the correct way to check the java versions?
ie. is it always inherited from the java version of the gradle process? or is there a toolchain/fork option we should be looking at?

JLLeitschuh: Good question. I don't think that, at this point, we are executing in the worker process. At this point we are still in the Gradle process, so it will be the JVM version of that process/classloader.

The workers are running in their own process/classloader, and static variables are unique-per-classloader hierarchy IIRC. As such, you can safely assume that any `JavaVersion.current()` calls will be accurate within the JVM classloader they are executing inside of.

tadfisher: Looking at Gradle's exec code, the default `JavaForkOptions` is set up [here](https://github.com/gradle/gradle/blob/ab01e7fc05e4d14186b0a706e0737fe0b20f458f/subprojects/core/src/main/java/org/gradle/process/internal/DefaultExecActionFactory.java#L369-L375):

```java
@Override
public JavaForkOptionsInternal newDecoratedJavaForkOptions() {
    JavaDebugOptions javaDebugOptions = objectFactory.newInstance(DefaultJavaDebugOptions.class, objectFactory);
    final DefaultJavaForkOptions forkOptions = instantiator.newInstance(DefaultJavaForkOptions.class, fileResolver, fileCollectionFactory, javaDebugOptions);
    forkOptions.setExecutable(Jvm.current().getJavaExecutable());
    return forkOptions;
}
```

So by default, Gradle appears to set the java executable for the child process to the same executable as the Gradle daemon. To override this, I believe we would need to call `setExecutable` ourselves with a path to some other Java toolchain's `javac`. Since we don't do that, `JavaVersion.current()` should be accurate for both the Gradle daemon process and the forked worker process.

wakingrufus:
As for testing, if this does fix #629 can you increase the maximum gradle version to the latest 8.x?
If not, this seems like a worthwhile fix, even if it is not the only thing needed for gradle 8 support

tadfisher:
Sure. I've been using this patch on Gradle 8, and I haven't seen any other showstoppers for Gradle 8 support.

IvanPizhenko:
@wakingrufus @JLLeitschuh  Hi, is there a chance for this to be merged and released anytime soon? We are also using ktlint and suffering from this issue.

JLLeitschuh:
CI currently fails for this PR which is currently a non-starter for merging this PR at this moment

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/testAnnotations.kt
```diff
@@ -1,21 +1,28 @@
...
 
+import java.io.File
+import java.util.stream.Stream
+import kotlin.streams.asStream
 import org.gradle.util.GradleVersion
 import org.jetbrains.kotlin.utils.addToStdlib.cast
 import org.jlleitschuh.gradle.ktlint.KtlintBasePlugin
 import org.junit.jupiter.api.extension.ExtensionContext
 import org.junit.jupiter.params.provider.Arguments
 import org.junit.jupiter.params.provider.ArgumentsProvider
```
tommyettinger, line 11: This currently is failing CI due to ```
/home/runner/work/ktlint-gradle/ktlint-gradle/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/testAnnotations.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
```
But, unless I forgot my ABCs, this already is in lexicographic order, and wasn't before!

Comment (IvanPizhenko) [reply plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/testAnnotations.kt:11]: Looks like "java" stuff should go **after** other things.
See example here: https://github.com/JLLeitschuh/ktlint-gradle/blob/main/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/AbstractPluginTest.kt

```kotlin

import org.jlleitschuh.gradle.ktlint.tasks.GenerateReportsTask
import org.jlleitschuh.gradle.ktlint.testdsl.GradleTestVersions
import org.junit.jupiter.api.io.TempDir
import java.io.File

@GradleTestVersions
abstract class AbstractPluginTest {
...
```

IvanPizhenko: not sure yet where "kotlin" should go, but will try to find an example.

IvanPizhenko: Found it! Example: https://github.com/JLLeitschuh/ktlint-gradle/blob/main/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/TaskConfigurationAvoidanceTest.kt

```kotlin

import org.jlleitschuh.gradle.ktlint.tasks.GenerateReportsTask
import org.jlleitschuh.gradle.ktlint.tasks.KtLintCheckTask
import org.jlleitschuh.gradle.ktlint.tasks.KtLintFormatTask
import org.jlleitschuh.gradle.ktlint.tasks.LoadReportersTask
import org.jlleitschuh.gradle.ktlint.testdsl.GradleArgumentsProvider
import org.jlleitschuh.gradle.ktlint.testdsl.GradleTestVersions
import org.jlleitschuh.gradle.ktlint.testdsl.build
import org.jlleitschuh.gradle.ktlint.testdsl.project
import org.junit.jupiter.api.DisplayName
import org.junit.jupiter.api.extension.ExtensionContext
import org.junit.jupiter.params.ParameterizedTest
import org.junit.jupiter.params.provider.Arguments
import org.junit.jupiter.params.provider.ArgumentsSource
import java.util.stream.Stream
import kotlin.streams.asStream

@GradleTestVersions
class TaskConfigurationAvoidanceTest : AbstractPluginTest() {
```

So Java & Kotlin should go separately **after** third-party stuff, without empty lines between third-party stuff and  Java & Kotlin stuff, but themselves, they must be ordered lexicographically. That's it.

IvanPizhenko: Anyone can fix this?

DmitriyZaitsev: @tommyettinger Of course, this is not an issue, but we should treat it as _something_ that violates the established rules of static analysis. This is the usual code hygiene, keeping the same style. I believe that it is better to politely ask the contributor to update the pull request than to appeal to the repository owner to merge anything just because we want it asap. :)

IvanPizhenko: > That explains it, I suppose, but doesn't really tell us why it's a build-stopping error... Doesn't IntelliJ IDEA reorder imports on its own rather often?

There are multiple approaches to code formatting in the project. If you don't care - then you don't care. If you care a bit, you use some linting tool, but allow its failures. If you strongly care - any formatting issue must be fixed, and that's what we see here. And this is not something unusual. Dor example, in my project we also use ktlint, and ktlint issue does fail the build too, because **we do care** about formatting of our code to that extent. So finally, to say true, I'm not interested in discussing too much the philosophy, whether it is right or wrong, it's up to project maintainers to define that. I'm most interested to have this merged and relased ASAP with whatever fixes must be done here to make that happen.

IvanPizhenko: And yes, default formatting in the IntelliJ may be diffrent from what is configured for ktlint, so at least you must have ktlint plugin for IntelliJ installed to be able to reformat accordingly or run something like `./gradlew ktlintFormat` in your pre-commit hook (but I would recommend to install the plugin).

tadfisher: Sorry about that! I've updated this.

tadfisher:
@JLLeitschuh Looks like CI is good to go. Let me know if the change to [TestVersions.maxSupportedKotlinVersion](https://github.com/JLLeitschuh/ktlint-gradle/pull/634/commits/df5cb9d1069339ce021d37fde44c647796396652#diff-64eca15e5797645ce6974873175f140eeffb5da5b6dd6e72cc879587e3f789a1R19-R25) makes sense; these are the versions listed in the Kotlin Gradle plugin documentation [here](https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin). Without that change, our minimum Gradle version (currently 6.8) will fail with Kotlin 1.8.10, which requires Gradle 6.8.3.

JLLeitschuh:
I'd like to wait for this if your fine with that

https://github.com/JLLeitschuh/ktlint-gradle/pull/642

wakingrufus:
LGTM. ill merge my other PR, then we can do this one

JLLeitschuh:
Just rebased the changes

wakingrufus (approved):

IvanPizhenko (approved):

Issue #633, opened 2023-02-20
additionalEditorconfigFile config has stopped working


hrules6872:
We had to upgrade the plugin to 11.2.0 to be able to use ktlint 0.48.2 but since then `additionalEditorconfigFile` has stopped working.
The plugin only pays attention to the `.editconfig` file located in the root of the project (where we were eventually forced to move it as a workaround).

wakingrufus:
Would you mind checking if it is working for you with ktlint 0.47.1? I suspect that this is specifically related to 0.48, and would like to confirm

hrules6872:
I just tried it and it still happens

wakingrufus:
OK, i just verified with ktlint upstream maintainer that this feature was removed in upstream ktlint 0.47+.
Would you mind explaining your usecase for this in more detail so that I can assess what kind of workarounds we might be able to provide in this plugin?

hrules6872:
TBH we only use this configuration to better organise the project so it may not be worth working on a solution, but it would be great to at least remove that option from the documentation :)

PR #636, opened 2023-02-22
Remove obsolete property

hrules6872:
closes https://github.com/JLLeitschuh/ktlint-gradle/issues/633


wakingrufus:
thanks for the PR!
We still support older versions of ktlint that do support this feature still, so I think instead of removing it, I will introduce a check for its usage in comvination with ktlint >=0.47, then we can remove it totally once our minimum ktlint version surpasses 0.47

wakingrufus:
Take a look at #637 instead

PR #638, opened 2023-02-23
Support disabledRules for ktlint version 0.48.x

matsudamper:
Using disabledRules gives this same warning.
https://github.com/JLLeitschuh/ktlint-gradle/issues/622

changed `disabled_rules` to `ktlint_standard_${ruleName}` for ktlint version 0.48.x.
https://github.com/pinterest/ktlint/commit/4d273d0f89bd8a36d27f17cc591634fb1bb971e4

I am not using `ktlint_experimental` and `custom-rule`. I have not implemented them because I cannot check the use case.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
@@ -156,6 +156,15 @@ abstract class KtLintWorkAction : WorkAction<KtLintWorkAction.KtLintWorkParamete
         if (disabledRules.isNotEmpty()) {
             userData["disabled_rules"] = disabledRules.joinToString(separator = ",")
         }
+        if (SemVer.parse(parameters.ktLintVersion.get()) < SemVer(0, 48, 0)) {
+            if (disabledRules.isNotEmpty()) {
+                userData["disabled_rules"] = disabledRules.joinToString(separator = ",")
+            }
+        } else {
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt
```diff
@@ -156,6 +156,15 @@ abstract class KtLintWorkAction : WorkAction<KtLintWorkAction.KtLintWorkParamete
         if (disabledRules.isNotEmpty()) {
             userData["disabled_rules"] = disabledRules.joinToString(separator = ",")
         }
+        if (SemVer.parse(parameters.ktLintVersion.get()) < SemVer(0, 48, 0)) {
+            if (disabledRules.isNotEmpty()) {
+                userData["disabled_rules"] = disabledRules.joinToString(separator = ",")
+            }
+        } else {
+            disabledRules.forEach {
+                userData["ktlint_standard_$it"] = "disabled"
```
wakingrufus, line 165: this is not safe, because not all rules are in the standard set.
there is really no way to know, just from the rule name, which set it is in.
supporting this would also require that plugin users include the full name with ruleset.
because of this, we are now recommending to just configure disabled rules using .editorconfig

matsudamper: I understood the policy of abolishing disabledRules and setting only with .editorconfig.

PR #650, opened 2023-03-03
fix release actions

wakingrufus:
VERSION_LATEST_RELEASE was getting updated too early
so the actual build to publish the plugin was trying to resolve the plugin that was not released yet


JLLeitschuh (approved):
LGTM!

PR #664, opened 2023-04-17
fix isRootEditorConfig

### LINKED ISSUE #575: `isRootEditorConfig()` method doesn't ever return `true`
👋 Hello to whoever will maintain the plugin :) 
I _think_ I found a bug where the `editorConfigRootRegex` doesn't work in the expected way. If I'm not mistaken, `lines` emitted by `useLines` won't ever emit a string with newline character at the end, so `\R` part of the regex is not going to be ever matched.

https://github.com/JLLeitschuh/ktlint-gradle/blob/ddd465e28d77b879384886e1eef5666ebe518b4d/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/PluginUtil.kt#L79

So if `.editorConfig` file contains
```editorconfig
root=true

[*.kt]
...
```
`useLines` will emit `"root=true"`, `""`, `"[*.kt]"]`..., where none of these strings will match the `^root\s?=\s?true\R` regex 👀 

image
if there was a newline character at the end - the regex would work as expected:
image

For more context: I was implementing a similar feature for [kotlinter-gradle](https://github.com/jeremymailen/kotlinter-gradle/pull/246), and I realized this doesn't look right

wakingrufus:
fixes #575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 -   fix new ktlint errors that come from our new default version of ktlint [#651](https://github.com/JLLeitschuh/ktlint-gradle/pull/651)
 -   fix syntax bug in release logic for VERSION_LATEST_RELEASE.txt [#651](https://github.com/JLLeitschuh/ktlint-gradle/pull/651)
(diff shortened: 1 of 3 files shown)


JLLeitschuh (approved):
LGTM! Thanks for fixing this!

PR #667, opened 2023-04-27
0.49 support

wakingrufus:
convert ktlint invocations from reflection to new method of using multiple compileOnly targets


JLLeitschuh:
Looking solid so far. I didn't realize how easy source sets were to work with these days. I would have created separate projects per version of Ktlint, but your solution is so much simpler. I love it!

Nice work! 🎉

plugin/build.gradle.kts
```diff
@@ -47,12 +47,96 @@ tasks.withType<PluginUnderTestMetadata>().configureEach {
...
+    add("adapterCompileOnly", libs.ktlint.core)
+    add("adapterImplementation", libs.semver)
+    add("adapter34Implementation", kotlin("reflect"))
+    add("adapter34CompileOnly", "com.pinterest.ktlint:ktlint-core:0.34.0")
+    add("adapter41CompileOnly", "com.pinterest.ktlint:ktlint-core:0.41.0")
+    add("adapter45CompileOnly", "com.pinterest.ktlint:ktlint-core:0.45.2")
+    add("adapter46CompileOnly", "com.pinterest.ktlint:ktlint-core:0.46.1")
+    add("adapter47CompileOnly", "com.pinterest.ktlint:ktlint-core:0.47.1")
+    add("adapter48CompileOnly", "com.pinterest.ktlint:ktlint-core:0.48.2")
+    //  add("adapter49CompileOnly", "com.pinterest.ktlint:ktlint-rule-engine-core:0.49.0")
```
JLLeitschuh, line 134: Do we need this still?

plugin/src/adapter/kotlin/org/jlleitschuh/gradle/ktlint/reporter/GenericReporterProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import java.io.PrintStream
+
+/**
+ * Abstraction over ReporterProvider and ReporterProviderV2
+ */
+interface GenericReporterProvider<T : GenericReporter<*>> {
+    val id: String
+    fun get(
+        out: PrintStream,
```
JLLeitschuh, line 11: You might want to consider renaming this variable as it conflicts with kotlin's `out` keyword.

Don't feel obligated to do this, but something to consider

plugin/src/adapter/kotlin/org/jlleitschuh/gradle/ktlint/worker/LintErrorResult.kt
```diff
@@ -10,9 +9,9 @@ import java.io.Serializable
  * @param lintedFile file that was checked by KtLint
  * @param lintErrors list of found errors and flag indicating if this error was corrected
  */
-internal data class LintErrorResult(
```
JLLeitschuh, line 13: I presume this is because kotlin doesn't see this class as being a part of the same project? There is a way to configure the kotlin compiler "friend path" but I don't know if that API is exposed via Gradle. So this is reasonable.

But if you're curious, feel free to look into it.

JLLeitschuh: Oh hey... Look what I found. A ticket I opened 5 years ago on this same topic: https://youtrack.jetbrains.com/issue/KT-20760

plugin/src/adapter34/kotlin/org/jlleitschuh/gradle/ktlint/reporter/SerializableReportersProviderLoader.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.io.ObjectInputStream
+import java.util.ServiceLoader
+
+class SerializableReportersProviderLoader :
+    ReportersLoaderAdapter<Reporter, SerializableReporterProvider, Ktlint34Reporter, Ktlint34ReporterProvider> {
+    override fun loadAllReporterProviders(): List<ReporterProviderWrapper<SerializableReporterProvider>> = ServiceLoader
+        .load(ReporterProvider::class.java)
+        .toList().map {
+            ReporterProviderWrapper(it.id, SerializableReporterProvider(it))
+        }
```
JLLeitschuh, line 15: Good use of service loaders IMHO!

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
```diff
@@ -0,0 +1,77 @@
...
+        .toList().map {
+            ReporterProviderWrapper(it.id, it)
+        }
+
+    override fun loadReporterProviders(serializedReporterProviders: File): List<ReporterProviderV2ReporterProvider> {
+        return ObjectInputStream(
+            serializedReporterProviders.inputStream().buffered()
+        ).use {
+            @Suppress("UNCHECKED_CAST")
+            it.readObject() as List<ReporterProviderV2<ReporterV2>>
```
JLLeitschuh, line 22: Sanity check. No way for this deserialization to come from untrusted user input? Arbitrary deserialization of untrusted java objects is a common source of remote code execution vulnerabilities, via deserialization gadget chains.

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
```diff
@@ -0,0 +1,77 @@
...
+                val provider = enabledProviders.find { reporterType.reporterName == it.id }
+                    ?: throw RuntimeException(
+                        "KtLint plugin failed to load reporter ${reporterType.reporterName}."
+                    )
+
+                val options = if (reporterType == ReporterType.PLAIN_GROUP_BY_FILE) {
+                    reporterType.options.associateWith { "true" }
+                } else {
+                    emptyMap()
+                }
```
JLLeitschuh, line 72: I feel like I've read this exact snippet a few times here. Any way to unify it into a common method?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBaselineSupportTest.kt
```diff
@@ -135,6 +135,7 @@ class KtlintBaselineSupportTest : AbstractPluginTest() {
                 buildGradle.appendText(
                     """
                     ktlint.version = "$ktLintVersion"
+                    ktlint.debug = true
```
JLLeitschuh, line 138: Very sane, why we haven't done this before, I don't know 😂

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/TestAnnotations.kt
```diff
@@ -12,7 +12,7 @@ import kotlin.streams.asStream
 
 object TestVersions {
     const val minSupportedGradleVersion = KtlintBasePlugin.LOWEST_SUPPORTED_GRADLE_VERSION
-    const val maxSupportedGradleVersion = "8.0.1"
+    const val maxSupportedGradleVersion = "8.1.1"
```
JLLeitschuh, line 15: Woot!

JLLeitschuh (commented):
In general, looking good. I would suggest using ValidatingInputStream when deserializing java objects if possible

plugin/build.gradle.kts
wakingrufus: no, ill remove

plugin/src/adapter/kotlin/org/jlleitschuh/gradle/ktlint/reporter/GenericReporterProvider.kt
wakingrufus: good call

plugin/src/adapter/kotlin/org/jlleitschuh/gradle/ktlint/worker/LintErrorResult.kt
wakingrufus: I was able to get friend paths working, but making this internal will also require KtLintInvocation to be internal, and then the work action can't invoke it without also being internal, and I assume those can't be internal

plugin/src/adapter34/kotlin/org/jlleitschuh/gradle/ktlint/reporter/SerializableReportersProviderLoader.kt
wakingrufus: I can't take credit, this code was already here :)

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
wakingrufus: Its the same logic, applied to different ReporterProvider types, and we do access properties on those types. We don't have duck typing so i'd have to wrap those classes and convert back and forth to commonize this code, and imho the duplication isn't bad enough to add that much complexity

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBaselineSupportTest.kt
wakingrufus: hah I needed this to debug some issues. I could remove it, but I guess we might as well leave this here otherwise we will probably forget about that option.

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
wakingrufus: its deserializing from a file created by this plugin, which is a list of the reporters used.
if you were able to replace that file between the time the plugin writes and and when it is read, I suppose it could be malicious input.  or if you modified the build to point `loadedReporterProviders` another location containing the bad file. This is always how this plugin has worked though (since i came to it). I think we could remove this serialization, but it would require doing the serviceloader call every time we need reporters which would impact performance.

plugin/src/adapter/kotlin/org/jlleitschuh/gradle/ktlint/worker/LintErrorResult.kt
JLLeitschuh: No worries

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
JLLeitschuh: There's also the validating deserializer that comes from apache Commons. I think we have a use of it elsewhere

plugin/src/adapter49/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ReportersProviderV2Loader.kt
JLLeitschuh: Fair enough! I thought that may have been what was probably going on. Just wanted to check

JLLeitschuh:
This is looking good. Anything else to make work?

JLLeitschuh (approved):
LGTM!

It may be a good sanity check to use the validating input stream when deserializing java objects though.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintClassesSerializer.kt
```diff
@@ -19,28 +17,17 @@ internal interface KtLintClassesSerializer {
...
-    ): List<ReporterProvider<*>>
-
     companion object {
-        fun create(ktLintVersion: SemVer): KtLintClassesSerializer =
-            if (ktLintVersion >= SemVer(0, 41, 0)) {
-                CurrentKtLintClassesSerializer()
-            } else {
-                OldKtLintClassesSerializer()
-            }
+        fun create(ktLintVersion: SemVer? = null): KtLintClassesSerializer =
```
JLLeitschuh, line 26: This argument doesn't look like it's used. Do we need it?

JLLeitschuh (approved):
One thing. Other than that. LGTM!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintClassesSerializer.kt
wakingrufus: fixed. thanks!

JLLeitschuh (approved):
LGTM!

Issue #576, opened 2022-04-05
Sometimes the plugin doesn't list all errors, only shows a partial list


jamesonwilliams:
In https://github.com/JLLeitschuh/ktlint-gradle/commit/e111faa23224e325c6170a3aa8b21283fe32eab9, I've made a few tweaks to the [`android-app` sample](https://github.com/JLLeitschuh/ktlint-gradle/tree/master/samples/android-app) to help demonstrate this issue.

--------

With those changes applied, the project will fail the `ktlintCheck` in `samples:android-app`, with the following errors.

When I run:
```
./gradlew samples:android-app:ktlintCheck
```

I get this output:
```
> Task :samples:android-app:ktlintMainSourceSetCheck FAILED
/Users/jameson.williams/ktlint-gradle/samples/android-app/src/main/java/org/jlleitschuh/gradle/ktlint/android_with_underscore/ItemDetailFragment.kt:1:1 Package name must not contain underscore (cannot be auto-corrected)
/Users/jameson.williams/ktlint-gradle/samples/android-app/src/main/java/org/jlleitschuh/gradle/ktlint/android_with_underscore/ItemDetailFragment.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between
```

Remark: two errors are shown, both in `ItemDetailFragment`. One error about package name, one error about import order.

-------

Now, let's suppose I fix the package name bug that my commit introduced, so there are no longer any `_` in the package names. (This is done in https://github.com/JLLeitschuh/ktlint-gradle/commit/6c2363e445480c863c5c2e775ad87df148cef40f.)

When I run:
```
./gradlew samples:android-app:ktlintCheck 
```

I get this output:
```
> Task :samples:android-app:ktlintMainSourceSetCheck FAILED
/Users/jameson.williams/ktlint-gradle/samples/android-app/src/main/java/org/jlleitschuh/gradle/ktlint/android/ItemDetailFragment.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between
/Users/jameson.williams/ktlint-gradle/samples/android-app/src/main/java/org/jlleitschuh/gradle/ktlint/android/ItemListActivity.kt:49:58 Unnecessary trailing comma before ")"
```

Result: Two errors are reported, one in `ItemDetailFragment` and a new one in `ItemListActivity`. The import order error is still present from before, but, _a new, previously-unreported error_ is uncovered in the `ItemListActivity`.

-------

[...]

Issue #491, opened 2021-06-01
Any support for non-incremental ktlintCheck?


abubics:
I've looked through the readme and open/closed issues, but can't find anything relevant :)

Is there any option to make ktlint not show incremental violations only? (short of also running `clean` (slow), or manually deleting the `build/intermediates/ktlint` directory / contents)

This behaviour is great for CI, where it will always start from a clean build, but locally if I run the lint check, fix some files manually, then run it again, it will pass while there are still outstanding issues.

The only other workflow I can think of is to always use the `ktlintFormat` task instead, which is ok (if inflexible) except where fixes aren't automatic.

gustavkarlsson:
The incremental nature of `ktlintCheck` seems like a very weird behavior in general. I thought it's a bug until I found the section in the README that explains it. Or am I misunderstanding this?

Isn't the whole point of any "check" that it's a pure function? I.e. produces the same result every time given the same inputs?

PR #672, opened 2023-06-05
add files previously found to have errors to incremental builds

wakingrufus:
add files previously found to have errors to the list of files to check in incremental builds. This will prevent the lint task from passing when errors are still present.

This should solve #491 and #576

### DIFF
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
@@ -494,6 +494,42 @@ class KtlintPluginTest : AbstractPluginTest() {
         }
     }
 
+    @DisplayName("Lint check should run incrementally and repeat errors")
+    @CommonTest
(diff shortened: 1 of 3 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -494,6 +494,42 @@ class KtlintPluginTest : AbstractPluginTest() {
...
+                val bar="foo"
+
+                """.trimIndent()
+            )
+
+            buildAndFail(CHECK_PARENT_TASK_NAME, "--info") {
+                assertThat(task(":$mainSourceSetCheckTaskName")?.outcome).isEqualTo(TaskOutcome.FAILED)
+                assertThat(output)
+                    .contains("Executing incrementally")
+                    .contains("Initial.kt")
```
wakingrufus, line 527: previously, Initial would not be listed (and if AnotherFile passed, the whole task would pass)

JLLeitschuh (approved):
LGTM!

PR #703, opened 2023-08-19
Fix for "Kotlin `sourceDir` isn't take into account"

### LINKED ISSUE #702: Android: `ktlint[SourceSetName]SourceSetCheck` task doesn't take kotlin `srcDir` into account
### Given:
Apply the plugin to an android project.
The android project does have a `main/src/kotlin` directory.
If you now run `ktlintMainSourceSetCheck` it will not run over this directory and therefore does not lint.

### Temporary solutions I found so far: 
* Move the source code to `src/main/java`.
* Add this code:
```kotlin
android.sourceSets.maybeCreate("main").java.srcDir("src/main/kotlin")
```

### Why isn't this working?
I guess (not 100% sure) I found the line in question:

https://github.com/JLLeitschuh/ktlint-gradle/blob/e5e60381f88f3eba9f3cc2904b7ae4a00b83d7da/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt#L70-L72

I guess this one will only create the expected task for the `sourceSet.java` source directories, but **not** for the `sourceSet.kotlin` source directories.

As you can see here, the [java property](https://developer.android.com/reference/tools/gradle-api/8.0/com/android/build/api/dsl/AndroidSourceSet#java()) is different to the [kotlin property](https://developer.android.com/reference/tools/gradle-api/8.0/com/android/build/api/dsl/AndroidSourceSet#kotlin()).

That would also explain why it would work when we add the kotlin (`src/main/kotlin`) `srcDir` to the `java` property.

If this is the case, a fix would be quite easy:
```diff
- val androidSourceSet = sourceSet.java as DefaultAndroidSourceDirectorySet
+ val androidSourceSet = sourceSet.kotlin as DefaultAndroidSourceDirectorySet
```

### Why would it solve it?
Because the `kotlin` property contains already the `main/src/java` source directory.

You can check this by running:
```groovy
android.sourceSets.all {
    println("Java -->" + (it.java as DefaultAndroidSourceDirectorySet).srcDirs)
}

android.sourceSets.all {
    println("Kotlin -->" + (it.kotlin as DefaultAndroidSourceDirectorySet).srcDirs)
}
```

This will print out something like:
```shell
Jaava -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTest/java]
Jaava -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestDebug/java]
Jaava -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestRelease/java]
Jaava -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/debug/java]
Jaava -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/main/java]

Kotlin -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTest/kotlin, /Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTest/java]
Kotlin -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestDebug/kotlin, /Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestDebug/java]
Kotlin -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestRelease/kotlin, /Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/androidTestRelease/java]
Kotlin -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/debug/kotlin, /Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/debug/java]
Kotlin -->[/Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/main/kotlin, /Users/stefan/Developer/Android/[project]/[path]/lib-[something]/src/main/java]
```

I'm ready to create an PR in case someone of the owners can confirm my assuptions 🙃

StefMa:
Closes #702 

### Description:

An detailed explanation *why* can be read in #702 

I also had the upgrade the AGP to version 7.0.0 because the 4.1.0 doesn't know the `kotlin` property 😁 
7.0.0 is the first version that supports hit.
Yes, this would be an potential breaking change.
But please take a look at the release dates for those plugins:
https://maven.google.com/web/index.html?q=tools.bui#com.android.tools.build:gradle:4.1.0 
-> Release at 10th December 2020

https://maven.google.com/web/index.html?q=tools.bui#com.android.tools.build:gradle:7.0.0
-> Release at 28th July 2021

As the Android community is normaly quite good in upgrading the *core* plugin to develop, I guess we are fine 😁 

### Changes:
* Upgrade Android Gradle Plugin from 4.1.0 to 7.0.0
* Use `kotlin` sourceSet instead of `java` sourceSet

### How to test:
1. Change the version in `/plugin/VERSION_CURRENT.txt` to something else e.g. "testKotlinSourceSet"
2. 
```shell
cd plugin 
./gradlew publishToMavenLocal
```
3. Include the `mavenLocal()` repository (to find that version)
4. Replace the current ktlint-plugin version (in your project) with the one you change din the `/plugin/VERSION_CURRENT.txt`

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt
@@ -1,14 +1,10 @@
 package org.jlleitschuh.gradle.ktlint.android
 
-import com.android.build.api.dsl.AndroidSourceSet
 import com.android.build.api.dsl.BuildFeatures
 import com.android.build.api.dsl.BuildType
 import com.android.build.api.dsl.CommonExtension
 import com.android.build.api.dsl.DefaultConfig
 import com.android.build.api.dsl.ProductFlavor
-import com.android.build.api.dsl.SigningConfig
-import com.android.build.api.variant.Variant
-import com.android.build.api.variant.VariantProperties
 import com.android.build.gradle.internal.api.DefaultAndroidSourceDirectorySet
 import org.gradle.api.Plugin
 import org.gradle.api.file.FileCollection
@@ -43,17 +39,8 @@ internal fun KtlintPlugin.PluginHolder.applyKtLintToAndroid(): (Plugin<in Any>) 
     }
 }
 
-@Suppress("UnstableApiUsage")
-private typealias AndroidCommonExtension = CommonExtension<
-    AndroidSourceSet,
-    BuildFeatures,
-    BuildType,
-    DefaultConfig,
-    ProductFlavor,
-    SigningConfig,
-    Variant<VariantProperties>,
-    VariantProperties
-    >
+private typealias AndroidCommonExtension =
+    CommonExtension<BuildFeatures, BuildType, DefaultConfig, ProductFlavor>
 
 /*
  * Variant manager returns all sources for variant,
@@ -69,7 +56,7 @@ private fun androidPluginConfigureAction(
 
         androidCommonExtension.sourceSets.all {
             // https://issuetracker.google.com/u/1/issues/170650362
-            val androidSourceSet = java as DefaultAndroidSourceDirectorySet
+            val androidSourceSet = kotlin as DefaultAndroidSourceDirectorySet
             // Passing Callable, so returned FileCollection, will lazy evaluate it
             // only when task will need it.
             // Solves the problem of having additional source dirs in
--- a/plugin/gradle/libs.versions.toml
+++ b/plugin/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 kotlin = "1.5.31"
 ktlint = "0.47.1" # last version prior to API changes. we can increase this more once we drop support for the old API.
-androidPlugin = "4.1.0"
+androidPlugin = "7.0.0"
 semver = "1.1.1"
 jgit = "5.6.0.201912101111-r"
 sl4fj = "1.7.30"


wakingrufus:
To be clear about what this breaks, it would mean that we now require AGP >=7.0 and Gradle >= 7.0?

I am ok with this, but I would like to queue  up some related maintenance work around dropping Gradle 6 support so we can merge them all around the same time and release it as 16.0

StefMa:
@wakingrufus did you had the time to look into this yet?

wakingrufus:
> @wakingrufus did you had the time to look into this yet? 

Not yet. I need to finish up https://github.com/JLLeitschuh/ktlint-gradle/pull/708 first

StefMa:
It's going to be happen 🤘 🙃

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt
```diff
@@ -69,7 +56,7 @@ private fun androidPluginConfigureAction(
 
         androidCommonExtension.sourceSets.all {
             // https://issuetracker.google.com/u/1/issues/170650362
-            val androidSourceSet = java as DefaultAndroidSourceDirectorySet
+            val androidSourceSet = sourceSet.kotlin as DefaultAndroidSourceDirectorySet
```
wakingrufus, line 59: this is a compile error currently. If you can fix this, I'd be happy to merge this PR.

JLLeitschuh:
> Yes, this would be an potential breaking change.

Do we want to accept this API breaking change? Does this break our guarantee of android plugin support range?

Do we currently have unit tests that cover the range of android plugins we want to ensure we support? If we don't we should.

StefMa:
I just saw that we set the AGP only as a `compileOnly` dependency
https://github.com/JLLeitschuh/ktlint-gradle/blob/a826dcdec179d971a8d3d91428214d177e262ff2/plugin/build.gradle.kts#L135
With that we can also workaround the breaking change by adding a try-catch 🤔 
```kotlin
val androidSourceSet = try {
  sourceSet.kotlin as DefaultAndroidSourceDirectorySet
} catch(_: java.lang.NoSuchMethodError) {
  java as DefaultAndroidSourceDirectorySet
} 
```

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/android/AndroidPluginsApplier.kt
StefMa: Fixed and rebased here https://github.com/JLLeitschuh/ktlint-gradle/pull/703/commits/eb47391571384f368901463c6d274e6bb3eb7b7b

Goooler:
Seems this could be closed for now, handled by #733.

PR #708, opened 2023-09-14
ktlint 1.0 support

### LINKED ISSUE #706: Support Ktlint 1.0
https://github.com/pinterest/ktlint/releases/tag/1.0.0

### LINKED ISSUE #707: Support `--disabled-rules` CLI option in `ktlintFormat` and `ktlintCheck`
Thinking of supporting the CLI option for disabling rules?
As disabling rules in `.editorconfig` is deprecated in KtLint version `0.48` and are removed in `0.49`, the only way to disable the rules globally without `.editorconfig` is using the `--disabled-rules` option via KtLint CLI.

ref) #267

wakingrufus:
Ktlint 1.0 support (fixes #706)

Allow editorconfig overrides in ktlint 0.49+ (fixes #707)  using https://github.com/pinterest/ktlint/pull/2194

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -88,6 +88,9 @@ sourceSets {
     val adapter50 by creating {
         compileClasspath += adapter.output
     }
+    val adapter100 by creating {
+        compileClasspath += adapter.output
+    }
     val adapters = listOf(
         adapter,
         adapter34,
@@ -97,7 +100,8 @@ sourceSets {
         adapter47,
         adapter48,
         adapter49,
-        adapter50
+        adapter50,
+        adapter100
     )
     val main by getting {
         kotlin {
@@ -121,7 +125,8 @@ val adapterSources = listOf(
     sourceSets.named("adapter47"),
     sourceSets.named("adapter48"),
     sourceSets.named("adapter49"),
-    sourceSets.named("adapter50")
+    sourceSets.named("adapter50"),
+    sourceSets.named("adapter100")
 )
 tasks.named<Jar>("shadowJar") {
     this.from(adapterSources.map { sourceSet -> sourceSet.map { it.output.classesDirs } })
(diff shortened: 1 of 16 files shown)


plugin/settings.gradle.kts
```diff
@@ -2,7 +2,7 @@ pluginManagement {
     val latestRelease = file("VERSION_LATEST_RELEASE.txt").readText().trim()
     plugins {
         id("org.jlleitschuh.gradle.ktlint") version latestRelease
-        id("org.jetbrains.kotlin.jvm") version "1.7.21"
+        id("org.jetbrains.kotlin.jvm") version "1.9.0"
```
wakingrufus, line 5: have to upgrade because ktlint did (to support context receivers)

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt
```diff
@@ -36,22 +37,31 @@ internal fun createKtlintConfiguration(target: Project, extension: KtlintExtensi
...
+                            target.dependencies.create("io.github.oshai:kotlin-logging:5.1.0")
+                        )
+                    }
+                }
+            )
+        )
     }
 
-private fun resolveGroup(ktlintVersion: String) = when {
-    SemVer.parse(ktlintVersion) < SemVer(0, 32, 0) -> "com.github.shyiko"
```
wakingrufus, line 48: we dont support 0.32 anymore

plugin/build.gradle.kts
```diff
@@ -88,6 +88,9 @@ sourceSets {
     val adapter50 by creating {
         compileClasspath += adapter.output
     }
+    val adapter1 by creating {
```
Goooler, line 91:
```suggestion
    val adapter100 by creating {
```

Goooler (approved):
We can note 

> Change default code style to ktlint_official

in changelog, if someone is upgrading from the old versions, may need to override the style like `ktlint_code_style = intellij_idea`.

JLLeitschuh (approved):
Looks good to me!

extmkv:
@wakingrufus @JLLeitschuh when are you planing to generate a new version?

plugin/build.gradle.kts
wakingrufus: good call. I made the change

wakingrufus:
> We can note
> 
> > Change default code style to ktlint_official
> 
> in changelog, if someone is upgrading from the old versions, may need to override the style like `ktlint_code_style = intellij_idea`.

This is only true based on which ktlint version you are selecting right? the default is still 0.47.1

Goooler:
Yeah, before 1.0.0 the default style is `intellij_idea` but `ktlint_official` for now, this is a potential change, which with so many format conflicts.

wakingrufus:
> Yeah, before 1.0.0 the default style is `intellij_idea` but `ktlint_official` for now, this is a potential change, which with so many format conflicts.

I think this will be important to note when we change the _default_ to 1.0, but for now, this will not be imposed on anyone who is not explicitly upgrading.  We also currently note:

> 
>  The version of ktlint used by default _may change_ between patch versions of this plugin.
> If you don't want to inherit these changes then make sure you lock your version here.

in the README.md as well, but I will improve that to point to the ktlint release notes for reference.

PR #712, opened 2023-09-22
fix "additionalEditorconfig not supported until ktlint 0.49" warning

wakingrufus:
fix "additionalEditorconfig not supported until ktlint 0.49" warning when additionalEditorconfig is not being used


JLLeitschuh (approved):
LGTM

Issue #579, opened 2022-04-22
Exclude configuration is ignored


jivimberg:
On plugin version `10.2.1` the following configuration is ignored:

```
ktlint {
    filter {
        exclude("**/generated/**")
    }
}
```

It works fine on version `10.2.0`.

Also tried with 

```
ktlint {
    filter {
        exclude { element -> element.file.path.contains("generated/") }
    }
}
```

And it yielded the same result.

Might be a regression from #266

dalewking:
Seeing this as well, but on any of the 10.x versions. Luckily in my case what I needed to exclude was an entire sourceset so was able to just disable the task.

trks1970:
Same here. For me, it did not work with 10.0.0 - 10.3.0

the workaround was to configure:
`filter {
            //exclude("**/generated/**")
            exclude { projectDir.toURI().relativize(it.file.toURI()).path.contains("/generated/") }
        }`

monosoul:
I had a similar issue in my project. In my case I was configuring src dirs _after_ project evaluation, so the sources were getting added after the filter was applied. At least that was my hypothesis, I didn't dig into the sources to confirm it. But what helped solving it - is to configure src dirs lazily at config time.

E.g.:

Before the change:
```kotlin
project.afterEvaluate {
    sourceSets.main {
        java {
            srcDirs(tasks.withType<CodegenTask>())
        }
    }
}
```

After the change:
```kotlin
project.sourceSets.main {
    java {
        srcDirs(tasks.withType<CodegenTask>())
    }
}
```

f4lco:
+1'd. Working with a file pattern is not working for me, too. I gave fixing this bug a shot, but working with Gradle's APIs gave me a headache.  I *think* this is the offending line:

https://github.com/JLLeitschuh/ktlint-gradle/blob/1b5b4e7c328e89a8486a0d1776b05d5d943513f8/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt#L113

Because we cannot pass [`getAsFileTree()`](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileCollection.html#getAsFileTree--) a root directory for the file tree, filtering by relative path, e.g., `**/generated/**`, does not work. We would need to work out a way to convert the `FileCollection` into a `FileTree` with a root, for example the project directory, which will be the base for relative path patterns to match against.

[...]

PR #715, opened 2023-09-29
Change "setSource", so it also applies the configured filter

aytchell:
In our project we're using gradle-ktlint plus we have a GraphQL API and generate stubs from a schema file.
We have team members working on Linux, MacOs and Windows.
To exclude the generated code from linting and formatting we've added this filter statement to our `build.gradle.kts`:

```
val sep = File.separatorChar
ktlint {
    version.set("0.49.1")
    filter {
        exclude { it.file.path.contains("${sep}generated$sep") }
    }
}
```

This works fine for our team members using Linux and MacOS.
But the exclude statement is ignored on the Windows machines (which looks a bit like https://github.com/JLLeitschuh/ktlint-gradle/issues/579 ).

The changes in this PR fix the exclude filter for our team members developing on Windows.


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
```diff
@@ -119,15 +120,29 @@ abstract class BaseKtLintCheckTask @Inject constructor(
...
+     * Filters from configuration will be applied
      *
      * @param source given source objects will be evaluated as per [org.gradle.api.Project.file].
      */
-    fun setSource(source: Any): BaseKtLintCheckTask {
-        sourceFiles = objectFactory.fileCollection().from(source)
+    fun setSource(source: FileTree): BaseKtLintCheckTask {
+        sourceFiles = objectFactory
+            .fileCollection()
+            .from({ source.matching(patternFilterable) })
```
wakingrufus, line 132: does this mean the filter is applied twice? since the provider on line 120 also applies the same filter?

wakingrufus:
Thanks for the PR! I have 1 question which I put on the change set itself.

wakingrufus:
FYI, out github actions run the full test suite on osx, linux, and windows, so you should be able to add a failing test and see it fail on the windows action (since this seems to be a windows-specific issue).

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/BaseKtLintCheckTask.kt
aytchell: Actually this is my first contact with gradle internals so I don't fully understand when these methods are called.
From diffing the debug output of Windows and Linux runs I got the impression, that filesystem changes (inotify et al triggered while generating the stubs) are somehow handled differently.

wakingrufus:
which gradle version are you using? this might be an issue that was fixed in gradle itself

aytchell:
I've managed to write a unit test which passes on Linux but fails on Windows.
The unit test has its own commit (514f7e1).
Should I modify this PR (or open another one) so that it only includes the new test? This way we could verify this behaviour with the PR test suite (I've only called `./gradlew build` on both OSes)

aytchell:
In the project we're using gradle 8.3

wakingrufus:
thanks for the failing test. the thing about your potential fix, is that it _shouldnt_ make a difference. the way apply the patternFilterable to the FileCollection, it is "live". as files are added to the collection, the filter will be applied to those new files. I wonder if maybe the windows failure is happening b/c the file is created too close to when the ktlint task runs, and maybe windows file system is not actually showing the file yet.

wakingrufus:
can you add that test to this branch so I can run the CI checks on it?

aytchell:
I've dropped the last commit (with the fix in it) from my remote branch, so we the only change is the new unit test.
Once the test suite ran (and "hopefully" failed) I can push the fix onto the branch and you can re-run the test suite

aytchell:
Hi there, is there anything missing from my side, so the test suite can start? If not: could anyone with the necessary privileges please start the tests?

aytchell:
I've rebased my branch to the latest `main` and fixed the finding of ktlint

JLLeitschuh:
Approving run now. Sorry about the delay

aytchell:
Ok, this is interesting. The new test fails on my Windows machine.
I'm gonna check if this is somehow timing related

PR #720, opened 2023-10-23
Drop support for older versions of gradle and ktlint

wakingrufus:
change default version of ktlint to 1.0.1
increase minimum supported gradle version to 7.4.1
 increase minimum supported ktlint version to 0.47.1
 support ktlint 1.0.1
remove retry plugin as it is included in enterprise plugin now
update code to utilize kotlin-dsl
fix ktlint errors
upgrade build to gradle 8 and add toolchain resolver
update reporter samples to ktlint 1.0.1


plugin/build.gradle.kts
```diff
@@ -235,13 +187,13 @@ tasks.withType<Test>().configureEach {
...
-            maxFailures.set(10)
-        }
-    }
+//    retry {
+//        val isCiServer = System.getenv().containsKey("CI")
+//        if (isCiServer) {
+//            maxRetries.set(2)
+//            maxFailures.set(10)
+//        }
+//    }
```
JLLeitschuh, line 196: If the plugin is bundled with the enterprise plugin, this should still work, right?

JLLeitschuh:
Does the supported Gradle version match the version range supported by the Android plugin?

plugin/build.gradle.kts
wakingrufus: yes, I forgot to re-enable. I will fix

wakingrufus:
> Does the supported Gradle version match the version range supported by the Android plugin?

We will be moving to requiring AGP 7.0 which requires Gradle 7.0, so this will be aligned.

JLLeitschuh (approved):
LGTM!

JLLeitschuh:
Nice work on this!

PR #722, opened 2023-10-30
Fix configuration cache for relative paths

### LINKED ISSUE #721: Configuration cache broken with relative set to true
Hello there.

When we set `relative.set(true)` our configuration cache breaks with something along the line of
```
855 problems were found storing the configuration cache, 285 of which seem unique.
- Task `:app:ktlintBenchmarkSourceSetFormat` of type `org.jlleitschuh.gradle.ktlint.tasks.GenerateReportsTask`: invocation of 'Task.project' at execution time is unsupported.
```

This is probably because
https://github.com/JLLeitschuh/ktlint-gradle/blob/431af6b838915abc874c63e92955c9272c5d08dd/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt#L144-L146
is using project directly.

According to the table on https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements project.rootDir should be replaced with `A task input or output property or a script variable to capture the result of using project.rootDir to calculate the actual parameter.`

Thanks in advance!

rschattauer:
resolves #721 

The new test would fail before the changes to generate reports task and succeed afterwards

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
@@ -90,6 +90,13 @@ abstract class GenerateReportsTask @Inject constructor(
     @get:Optional
     internal abstract val baseline: RegularFileProperty
 
+    /**
+     * Reading a project's rootDir within a task's action is not allowed for configuration
+     * cache, so read it eagerly on task initialization.
+     * see: https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache
+     */
+    private val rootDir: File = project.rootDir
+
     init {
         // Workaround for https://github.com/gradle/gradle/issues/2919
         onlyIf {
@@ -142,7 +149,7 @@ abstract class GenerateReportsTask @Inject constructor(
                 param.baseline.set(baseline)
                 param.projectDirectory.set(projectLayout.projectDirectory)
                 if (relative.get()) {
-                    param.filePathsRelativeTo.set(project.rootDir)
+                    param.filePathsRelativeTo.set(rootDir)
                 }
             }
         }
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/ConfigurationCacheTest.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/ConfigurationCacheTest.kt
@@ -13,7 +13,6 @@ import org.junit.jupiter.api.DisplayName
 @GradleTestVersions
 class ConfigurationCacheTest : AbstractPluginTest() {
     private val configurationCacheFlag = "--configuration-cache"
-    private val configurationCacheWarnFlag = "--configuration-cache-problems=warn"
 
     @DisplayName("Should support configuration cache without errors on running linting")
     @CommonTest
@@ -29,7 +28,6 @@ class ConfigurationCacheTest : AbstractPluginTest() {
 
             build(
                 configurationCacheFlag,
-                configurationCacheWarnFlag,
                 CHECK_PARENT_TASK_NAME
             ) {
                 assertThat(task(":$mainSourceSetCheckTaskName")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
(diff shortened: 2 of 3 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/ConfigurationCacheTest.kt
```diff
@@ -58,15 +59,13 @@ class ConfigurationCacheTest : AbstractPluginTest() {
             val formatTaskName = KtLintFormatTask.buildTaskNameForSourceSet("main")
             build(
                 configurationCacheFlag,
-                configurationCacheWarnFlag,
```
JLLeitschuh, line 61: Can you elaborate on why this isn't needed anymore?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
```diff
@@ -90,6 +90,8 @@ abstract class GenerateReportsTask @Inject constructor(
     @get:Optional
     internal abstract val baseline: RegularFileProperty
 
+    private val rootDir: File = project.rootDir
```
JLLeitschuh, line 93: Does this need to be class scoped? Or can it be scoped smaller, to only the `init` block? Also, can you elaborate on why this fixes this, or, at minimum, can you include an explicit comment in the code explaining why this is extracted to a variable eagerly. That way the rationale for this change isn't lost in the future? 🙂

JLLeitschuh (commented):
Just a few bits of feedback and some questions. Other than that, nice work!

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
rschattauer: According to https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:~:text=Use%3A-,project.rootDir,-A%20task%20input for configuration caching rootDir you should use
`A task input or output property or a script variable to capture the result of using project.rootDir to calculate the actual parameter.`
So according to https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution `A task must not use any Project objects at execution time. This includes calling Task.getProject() while the task is running.` and to achieve that I was calling it within the class. I could also let the input set the relative rootdir path, but I thought this is easier to grasp? I added a comment to hopefully make it understandable.

What do you mean with init block scoped? There is no difference to this and 
```
private val rootDir: File
init {
     rootDir = project.rootDir
}
``
`
or do I misunderstand you?

File: plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/ConfigurationCacheTest.kt (hunk 1 above)
Comment (rschattauer) [reply plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/ConfigurationCacheTest.kt:61]: I removed it everywhere in this test now. Reason being that this would not fail the test when configuration actually fails. With relative path you would get a message that reads something like
```
FAILURE: Build failed with an exception.

* What went wrong:
Configuration cache problems found in this build.

2 problems were found storing the configuration cache, 1 of which seems unique.
```
So we probably want this test to fail in the future if someone implements something that is not capable of configuration caching?

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/tasks/GenerateReportsTask.kt
JLLeitschuh: Sorry, I misunderstood the code. I thought the `rootDir` was only used inside of the `init` block, but re-reading the code, I was completely wrong. My mistake.

Thanks for adding a comment inline in the code! 🙂

JLLeitschuh (approved):
LGTM!

JLLeitschuh:
Please add a changelog entry for this change. Then this should be good to go

rschattauer:
Added a line to the changelog @JLLeitschuh

JLLeitschuh (approved):
LGTM! Now to wait for CI

rschattauer:
@JLLeitschuh , I don't have to click/do anything to continue, correct?

JLLeitschuh:
Merged! Thanks for the fix!!

PR #725, opened 2023-11-28
update configure-pagefile-action task

wakingrufus:
update VERSION_LATEST_RELEASE.txt manually


.github/workflows/build-and-check.yml
```diff
@@ -99,8 +99,10 @@ jobs:
       - name: set up JDK 1.8
         uses: actions/setup-java@v2
         with:
-          java-version: 8
+          java-version: 11
```
wakingrufus, line 102: might as well run this on java 11

.github/workflows/build-and-check.yml
```diff
@@ -99,8 +99,10 @@ jobs:
...
           distribution: 'zulu'
+      - name: Install lintian
+        run: sudo apt install -qq lintian=2.114.0ubuntu1.3
```
wakingrufus, line 105: checkbashisms was resolving its transitive dependency on lintian to a version that no longer exists. i fixed it by explicitly installing a version of the dependency that actually exists

plugin/build.gradle.kts
```diff
@@ -306,10 +306,12 @@ gradlePlugin {
         register("ktlintPlugin") {
             id = "org.jlleitschuh.gradle.ktlint"
             implementationClass = "org.jlleitschuh.gradle.ktlint.KtlintPlugin"
+            displayName = "Ktlint Gradle Plugin"
```
wakingrufus, line 309: to pass new validation rules (this caused previous release to fail)

.github/workflows/build-and-check.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: configure windows pagefile
         if: ${{ matrix.os  == 'windows-latest' }}
-        uses: al-cheb/configure-pagefile-action@v1.2
+        uses: al-cheb/configure-pagefile-action@v1.3
```
wakingrufus, line 33: to solve node version warning

.github/workflows/build-and-check.yml
JLLeitschuh: Sure... But then change the title of the step too 🙂😂

JLLeitschuh:
One thing to fix, then happy to merge! 🙂

PR #726, opened 2023-11-30
remove KtLintIdea Plugin

wakingrufus:
This plugin is no longer needed as ktlint configuration is driven by .editorconfig now, which idea will respect out of the box

The plugin was already removed in 12.0.0, but we need to remove the declaration of the plugin in the build scripts to pass validation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


Spikhalskiy:
This section of Readme should probably be deleted: https://github.com/JLLeitschuh/ktlint-gradle#intellij-idea-only-plugin

rtsketo:
Shouldn't the README.md be updated as well?

It says:
```
plugins {
  id("org.jlleitschuh.gradle.ktlint-idea") version "<current_version>"
}
```

Honest question, otherwise I would have created a PR for this.

JLLeitschuh:
Yes, please do feel free to create a PR to update the README 😄

PR #732, opened 2023-12-08
fix: apply configuration for source sets and targets that are added after the plugin is applied

odzhychko:
Use `DomainObjectCollection.all` instead of `Iterable<T>.forEach`. `all` executes the given function also against any objects subsequently added objects

The problem was introduced here
https://github.com/JLLeitschuh/ktlint-gradle/commit/a547de5601caec50ed827b1a5fe6c8207d41d218?diff=unified&w=0#diff-51eee411b698815c113da41ed3e9f5c31ebc7cb926df3580835df154cdd721e0L45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


wakingrufus:
LGTM! Thanks for the fix!

PR #758, opened 2024-04-02
Fix additionalEditorconfig property not being accounted for up-to-date checks and caching

### LINKED ISSUE #750: Caches work incorrectly: additionalEditorconfig should be @Input instead of @Internal
Steps to reproduce:

1. Write some code with ktlint violations
2. Run ktlint via plugin with enabled build cache
3. Disable the violated rule using `additionalEditorconfig`
4. Run ktlint again with enabled build cache

Expected behaviour: ktlint will be re-run with updated editorconfig and will not report the violations

Actual behaviour: ktlint will use the cached result

As far as I understand, the cause of this is that `BaseKtLintCheckTask.additionalEditorconfig` is not checked when computing cache keys for the task because it is marked as `@Internal`, and should be `@Input` instead.

Unless there are any reasons for this property being `@Internal`, I would gladly submit a PR changing it to `@Input`.

m-burst:
Fixes #750

### DIFF
--- a/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
+++ b/plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
@@ -147,6 +147,26 @@ class EditorConfigTests : AbstractPluginTest() {
         }
     }
 
+    @DisplayName("Check task should rerun if additionalEditorconfig property changes")
+    @CommonTest
+    fun checkRerunOnAdditionalEditorconfigPropertyChange(gradleVersion: GradleVersion) {
+        project(gradleVersion) {
+            withAdditionalEditorconfigProperty(120)
+            withCleanSources()
+
+            build(CHECK_PARENT_TASK_NAME) {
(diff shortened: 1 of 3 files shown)


plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/EditorConfigTests.kt
```diff
@@ -147,6 +147,26 @@ class EditorConfigTests : AbstractPluginTest() {
         }
     }
 
+    @DisplayName("Check task should rerun if additionalEditorconfig property changes")
+    @CommonTest
+    fun checkRerunOnAdditionalEditorconfigPropertyChange(gradleVersion: GradleVersion) {
```
m-burst, line 152: I made sure that the test fails if `val additionalEditorconfig` is `@Internal` before changing it to `@Input`

wakingrufus (approved):
LGTM Thanks!

wakingrufus:
You will need to update the changelog to pass checks as well

m-burst:
Updated the changelog

JLLeitschuh (approved):
Thank you for fix, and the unit test, both are greatly appreciated! TDD for the win!

PR #763, opened 2024-05-01
update versions for testing

wakingrufus:
update max gradle to 8.6
update ktlint versions to be latest patch version of each minor version supported (0.47-1.2) update max AGP version to 8.4
update max kotlin version to 1.9.23


JLLeitschuh:
Sorry I never got back to you, thanks for merging and releasing. 🎉

PR #802, opened 2024-11-11
Update commons-io to 2.17.0

jowiho:
Update the `commons-io` library to the latest version to mitigate CVE-2024-47554. Though I don't believe that this CVE can be exploited through ktlint-gradle, it's still a good practice to avoid CVEs, if only to stop automated scanners from flagging this (as a false positive).


JLLeitschuh:
Hi @jowiho!

Thanks for your contribution! This will require an update to our https://github.com/JLLeitschuh/ktlint-gradle/blob/main/CHANGELOG.md. Would you be so kind as to add a line to that doc? Other than that, I'm more than happy to merge this change!

Thank you so much!

jowiho:
@JLLeitschuh, sure! I've added a commit to update `CHANGELOG.md`. Thanks for the reminder.

JLLeitschuh (approved):
LGTM

PR #819, opened 2024-12-10
update gradle wrapper to 8.11.1

wakingrufus:
add gradle api version metadata to plugin artifact
update max gradle version for testing to 8.11.1


plugin/build.gradle.kts
```diff
@@ -306,6 +306,12 @@ fun setupPublishingEnvironment() {
 
 setupPublishingEnvironment()
 
+configurations.named("implementation") {
+    attributes {
+        attribute(GradlePluginApiVersion.GRADLE_PLUGIN_API_VERSION_ATTRIBUTE, objects.named("7.4"))
```
JLLeitschuh, line 311: What is this? Also, why is this version hard coded? What is it related to?

wakingrufus: It's a way of declaring gradle module metadata to express the minimum gradle API needed. Its similar to how gradle detects when you try to use a library compiled on jdk11 in a jdk8 project. https://docs.gradle.org/current/userguide/variant_attributes.html#sec:gradle-plugins-default-attributes

This way, we don't need to programmatically check the gradle version. The version i selected here aligns with the minimum gradle version we support and test against

PR #833, opened 2025-02-14
test against newest kotlin, gradle, ktlint, and AGP versions

wakingrufus:
drop testing for gradle 7.4 and 7.5 because they cause our tests with kotlin 2.x to fail, but the plugin will still work in those gradle versions if kotlin 1.x is being used


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
```diff
@@ -54,20 +50,7 @@ open class KtlintBasePlugin : Plugin<Project> {
         )
     }
 
-    /**
-     * @deprecated Now that we declare gradle API metadata, this code should not be needed.
-     * Ee need to check which version of gradle introduced gradle API metadata checking
-     */
-    @Deprecated("Now that we declare gradle API metadata, this code should not be needed")
-    private fun Project.checkMinimalSupportedGradleVersion() {
```
JLLeitschuh, line 62: From what version is this metadata standard supported from?
Are we sure we want to yank this out? It seems like a reasonable sanity check if someone does something silly and applies the plugin to an ancient version of Gradle

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 - Update Gradle wrapper and Gradle versions for testing. [#819](https://github.com/JLLeitschuh/ktlint-gradle/pull/819)
+- test against newest kotlin, gradle, ktlint, and AGP versions [#833](https://github.com/JLLeitschuh/ktlint-gradle/pull/833). Drop testing for gradle 7.4 and 7.5 because they cause our tests with kotlin 2.x to fail, but the plugin will still work in those gradle versions if kotlin 1.x is being used
```
JLLeitschuh, line 11:
```suggestion
- Test against newest kotlin, gradle, ktlint, and AGP versions [#833](https://github.com/JLLeitschuh/ktlint-gradle/pull/833). Drop testing for gradle 7.4 and 7.5 because they cause our tests with kotlin 2.x to fail, but the plugin will still work in those gradle versions if kotlin 1.x is being used
```

JLLeitschuh (commented):
LGTM, just a few small things

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt
wakingrufus: Actually the reason I chose to yank this now is that any version of gradle lower than 7.6 + kotlin 2.x causes the build to fail in a different built-in way (with a nice message). So i stopped testing against 7.4. But it still works as long as you dont use kotlin 2 x, so I left the api compatibility declaration at 7.4.
But to test this method now, we would need a specific test that uses gradle 7.4 and kotlin 1.x 
I can add that and reinstate this function.

JLLeitschuh: Please do, if you have time and it's not too much of a headache

PR #837, opened 2025-02-25
Fix dependency issue for git hook tasks

ntkoopman:
When running one of the Git hook tasks together with another task, Gradle will complain with an error:

> Reason: Task ':addKtlintCheckGitPreCommitHook' uses this output of task ':samples:kotlin-reporter-using:compileKotlin' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

This is caused by `@InputDirectory` marking the entire project root and contents as input, so any task that modifies any file in the project will cause Gradle to add an implicit dependency.

You can see this behaviour by running `./gradlew compileKotlin addKtlintCheckGitPreCommitHook`.

Build result on the first commit of this PR, to show it failing: https://github.com/ntkoopman/ktlint-gradle/actions/runs/13514290368/job/37760683637

This is the same issue as https://github.com/JLLeitschuh/ktlint-gradle/pull/545, except this PR keeps the inputs.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
@@ -162,6 +162,11 @@ open class KtlintInstallGitHookTask @Inject constructor(
 
     @TaskAction
     fun installHook() {
+        /*
+         * Note: `projectDir` and `rootDirectory` are both using `Property<String>` instead
+         * of a `DirectoryProperty` because we want to rely upon the *path* not the *contents*
+         * https://github.com/JLLeitschuh/ktlint-gradle/pull/837
+         */
         val repo = RepositoryBuilder().findGitDir(File(projectDir.get())).setMustExist(false).build()
         if (!repo.objectDatabase.exists()) {
             logger.warn("No git folder was found!")
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
@@ -4,15 +4,14 @@ package org.jlleitschuh.gradle.ktlint
 
 import org.eclipse.jgit.lib.RepositoryBuilder
 import org.gradle.api.DefaultTask
-import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.ProjectLayout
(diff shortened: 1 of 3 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -151,19 +150,19 @@ open class KtlintInstallGitHookTask @Inject constructor(
...
-        set(projectLayout.projectDirectory)
+    @get:Input
+    internal val projectDir: Property<String> = objectFactory.property(String::class.java).apply {
+        set(projectLayout.projectDirectory.asFile.absolutePath)
     }
 
-    @get:InputDirectory
-    internal val rootDirectory: DirectoryProperty = objectFactory.directoryProperty().apply {
-        set(project.rootDir)
+    @get:Input
```
wakingrufus, line 158: this is the correct way to make an input depend on the _path_ and not the _contents_ :+1:

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/GitHook.kt
```diff
@@ -151,19 +150,19 @@ open class KtlintInstallGitHookTask @Inject constructor(
...
+    internal val rootDirectory: Property<String> = objectFactory.property(String::class.java).apply {
+        set(project.rootDir.absolutePath)
     }
 
     @TaskAction
     fun installHook() {
-        val repo = RepositoryBuilder().findGitDir(projectDir.get().asFile).setMustExist(false).build()
```
wakingrufus, line 166: I think we have some work to do in terms of properly making this task depend on the _git directory_ as an input. But i think this PR is fine for now and we can refine later

wakingrufus:
Thanks for the contribution! please update the changelog file (see the failed check for this for details). I am approving the ci checks to run, and once  everything passes, I will open to merging this

JLLeitschuh:
Hey @ntkoopman in the future, please don't create your pull request directly from your `main` branch. It makes it impossible for maintainers like myself to push changes to your branch which makes it more difficult for us to merge your changes.

Would you be willing to add a `CHANGELOG.md` file change with the following contents under `Unreleased`:

```markdown
### Fixed

- Fix `addKtlintCheckGitPreCommitHook` task marking the entire project root and contents as input ([#837](https://github.com/JLLeitschuh/ktlint-gradle/pull/837))
```

PR #865, opened 2025-04-25
Use manual build mode for java-kotlin in `codeql.yml`

redsun82:


.github/workflows/codeql.yml
```diff
@@ -73,17 +77,11 @@ jobs:
...
-      - name: Run CodeQL Autobuild
+      - name: Build
         # Now that an explicit java version has been set, use it
         if: matrix.language == 'java-kotlin'
-        uses: github/codeql-action/autobuild@v3
-        with:
-          # Only scan the plugin directory as the rest of the code is just sample code
-          working-directory: plugin
+        run: |
+          ./gradlew -p plugin --no-daemon --no-build-cache --console=plain --no-configuration-cache -x test -x check
```
JLLeitschuh, line 84: This won't inject the init script that the autobuilder uses though. Don't we want the init script to configure Gradle correctly?

redsun82: calling `github/codeql-action/init` changes the environment variables in a way that anything done in between it and the finalize action will be analyzed by CodeQL. Technically it uses LD_PRELOAD on Linux, and INSERT_DYLD_LIBRARIES on macOS. On windows there is another slightly different mechanism, and on macOS and Windows we do require running commands under a special runner, but on Linux it just works.

JLLeitschuh: I added a bit more to the concerns I have with this solution here: 

https://github.com/github/codeql/issues/19374#issuecomment-2830737377

I'm willing to accept this solution for now, but I think long-term this is masking a larger problem and is leaving build speed on the ground because of how the build-cache is entirely disabled

redsun82: ah, I see what you meant now, that's a gradle init script I wasn't aware of.

Issue #815, opened 2024-12-05
Apply ktlint 1.5.0 to project


doljae:
# Summary

Please apply [ktlint 1.5.0](https://github.com/pinterest/ktlint/releases/tag/1.5.0) to project

# Description

With the recent release of [Kotlin 2.1.0](https://github.com/jetbrains/kotlin/releases), there seems to be a small problem with [compatibility issues](https://github.com/pinterest/ktlint/issues/2882) with this across the open source for the Kotlin linter.

The `ktlint` has a new release that is shelved for Kotlin 2.1.0, so hopefully this will be quickly reflected in the project so that we can use the latest version of Kotlin and `ktlint-gradle` together 🙂

mezpahlan:
Does this work for you?

```
configure<KtlintExtension> {
  version.set("1.5.0")
}
```

doljae:
Hello @mezpahlan 🙂 
With the configuration from your comment, seems the plugin use `ktlint 1.5.0` internally.

I will use this way until the full version of the plugin is upgraded, thank you!

Issue #809, opened 2024-11-29
Error with kotlin 2.0.1


nckdistributable:
Hi, I'm getting an error when executing the ktlintCheck task. The error is:

Class org.jetbrains.kotlin.lexer.KtTokens does not have member field 'org.jetbrains.kotlin.lexer.KtModifierKeywordToken HEADER_KEYWORD'

I found that this keyword was removed in the most recent versions of the Kotlin compiler. Is the current version of ktlint aware of this?

Thanks!

gmmoreira:
This error has been fixed in ktlint lib version 1.4.0, but this plugin is using an slightly older version of the lib.

In build.gradle.kts we can configure the plugin to use the most recent version of ktlint.

```
ktlint {
    version.set("1.4.1")
}
```

That should fix the error for now.

JLLeitschuh:
A PR updating the default Ktlint version would be appreciated and accepted with a simple unit test and a CHANGELOG entry

CPColin:
I hope this PR is what you were looking for: #818

dumitrurogojinaru:
setting to ktlint version to 1.5 leads to failing on a lot of validations, did anyone experienced the same?

wakingrufus:
> setting to ktlint version to 1.5 leads to failing on a lot of validations, did anyone experienced the same?

if you were p reviously on a pre-1.0 version of ktlint, then yeah a lot of things changed in ktlint. This is a big reason why we have been dragging our feet on updating the default version in ktlint-gradle. I recommend looking at the ktlint github for advise on the changes to the ktlint rules

mkubele:
This is still valid issue for me with v13.0.0 and Kotlin v2.2.0

PR #889, opened 2025-05-22
remove support for ktlint < 1

wakingrufus:
remove support for ktlint < 1
increase minimum kotlin version to 1.7 because less than this is deprecated by KGP and support will be removed soon
update default ktlint version to 1.5 in order to avoid unnecessary bug reports such as https://github.com/JLLeitschuh/ktlint-gradle/issues/815 and https://github.com/JLLeitschuh/ktlint-gradle/issues/809


.editorconfig
```diff
@@ -11,6 +11,9 @@ ij_kotlin_packages_to_use_import_on_demand = unset
 ij_kotlin_name_count_to_use_star_import = 999
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ktlint_code_style=intellij_idea
+ktlint_standard_function-expression-body = disabled
+ktlint_standard_function-signature = disabled
+ktlint_standard_class-signature = disabled
```
JLLeitschuh, line 16: Why disable these? Do you want to enable them in a follow up cleanup PR?

plugin/build.gradle.kts
```diff
@@ -179,10 +146,8 @@ kotlin {
 tasks.withType<Test> {
     dependsOn("publishToMavenLocal")
     useJUnitPlatform()
-    maxParallelForks = (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
-    doFirst {
-        logger.lifecycle("maxParallelForks for '$path' is $maxParallelForks")
-    }
+    maxParallelForks = 6 // no point in this being higher than the number of gradle workers
```
JLLeitschuh, line 149: Can you explain this a bit more?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtLintSupportedVersionsTest.kt
```diff
@@ -173,17 +173,9 @@ class KtLintSupportedVersionsTest : AbstractPluginTest() {
...
-            "0.50.0",
-            "1.0.1",
-            "1.1.1",
-            "1.2.1",
-            "1.3.1",
-            "1.4.1",
-            "1.5.0"
+            "1.0.1", // lowest
+            "1.5.0", // default
+            "1.6.0" // highest
```
JLLeitschuh, line 178: Should we keep a few more of these in here?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -599,15 +565,16 @@ class KtlintPluginTest : AbstractPluginTest() {
...
-                    $KTLINT_RULESET_CONFIGURATION_NAME "com.pinterest.ktlint:ktlint-cli-ruleset-core:1.0.0"
-                }
-                """.trimIndent()
+ktlint {
+    version = "1.0.1"
+}
+dependencies {
+    $KTLINT_RULESET_CONFIGURATION_NAME "com.pinterest.ktlint:ktlint-cli-ruleset-core:1.0.0"
+}
+"""
```
JLLeitschuh, line 577: Formatting?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -625,11 +592,13 @@ class KtlintPluginTest : AbstractPluginTest() {
...
-                    $KTLINT_REPORTER_CONFIGURATION_NAME "com.pinterest.ktlint:ktlint-cli-ruleset-core:1.0.0"
-                }
-                """.trimIndent()
+ktlint {
+    version = "1.0.1"
+}
+dependencies {
+    $KTLINT_REPORTER_CONFIGURATION_NAME "com.pinterest.ktlint:ktlint-cli-ruleset-core:1.0.0"
+}
+"""
```
JLLeitschuh, line 601: Formatting?

plugin/build.gradle.kts
wakingrufus: when junit forks processes to run parallel tests, it dispatches each one to a gradle worker. On my machine, this calculation was resulting in about 11 forks, but since we only had 4 workers, only 4 would run at a time. i increased workers to 6, and then hard coded this to match. we may want to reduce the test forks even more if we want other gradle tasks to execute while tests are running

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtLintSupportedVersionsTest.kt
wakingrufus: the post-1.0 API and behavior has been very stable, so it seemed a bit redundant to test _every_ minor version, but this is just a build performance improvement. I have no problem with adding more. what do you think?  all of them again, or just a bigger subset?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
wakingrufus: this is another valid way of formatting text blocks. I prefer it as the other way tends to seem like it has an arbitrary indentation level, and this way does not require a trim function

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/AndroidSetup.kt
```diff
@@ -66,8 +66,8 @@ fun androidProjectSetup(
         |    }
         |
         |    plugins {
-        |        id("org.jetbrains.kotlin.android") version("$kotlinPluginVersion")
-        |        id("org.jlleitschuh.gradle.ktlint") version("${TestVersions.pluginVersion}")
+        |        id("org.jetbrains.kotlin.android") version ("$kotlinPluginVersion")
+        |        id("org.jlleitschuh.gradle.ktlint") version ("${TestVersions.pluginVersion}")
```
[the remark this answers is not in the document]
wakingrufus: the space was added to satisfy a new ktlint rule

PR #927, opened 2025-08-03
Integration & Testing of the Problems API

vanessamj99:
A new file was created to report ktlint issues to the problems API to act as another option of error reporting. A testing file was also added to test the functionality of the new integration using mockito.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -118,6 +118,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle.plugin)
     compileOnly(libs.android.gradle.plugin)
     compileOnly(kotlin("stdlib-jdk8"))
+    compileOnly(gradleApi())
     shadowImplementation(libs.semver)
     shadowImplementation(libs.jgit)
     shadowImplementation(libs.commons.io)
@@ -129,6 +130,9 @@ dependencies {
     testImplementation(libs.kotlin.reflect)
     testImplementation(libs.ktlint.rule.engine)
     testImplementation(libs.archunit.junit5)
+    testImplementation(gradleApi())
+    // Used to test the problems API
+    testImplementation("org.mockito.kotlin:mockito-kotlin:4.1.0")
 }
 
 kotlin {
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ProblemsApiReporter.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ProblemsApiReporter.kt
@@ -1,0 +1,57 @@
+package org.jlleitschuh.gradle.ktlint.reporter
+
+import org.gradle.api.Incubating
+import org.gradle.api.problems.ProblemGroup
+import org.gradle.api.problems.ProblemId
+import org.gradle.api.problems.ProblemReporter
+import org.gradle.api.problems.Problems
+import org.gradle.api.problems.Severity
+import org.jlleitschuh.gradle.ktlint.worker.SerializableLintError
+import javax.inject.Inject
+
+@Incubating
+class ProblemsApiReporter {
+
+    private var problems: Problems? = null
+
+    @Inject
+    public constructor(problems: Problems) {
+        this.problems = problems
+    }
+
+    public constructor() {
+        this.problems = null
+    }
+
+    fun reportProblems(lintErrors: Map<String, List<SerializableLintError>>) {
+        val reporter: ProblemReporter? = problems?.reporter
+
+        lintErrors.forEach { (filePath, errors) ->
+            errors.forEach { error ->
+                val group = ProblemGroup.create("validation", "ktlint issue")
+                val id = ProblemId.create(error.ruleId, error.detail, group)
+                reporter?.report(id) {
+                    fileLocation(filePath)
+                    lineInFileLocation(filePath, error.line)
+                    details(error.detail)
+                    severity(Severity.WARNING)
+                    solution("Run ktlintFormat to auto-fix this issue")
+                }
+            }
+        }
+    }
+
+    fun reportProblem(error: SerializableLintError, filePath: String) {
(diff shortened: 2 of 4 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ProblemsApiReporter.kt
```diff
@@ -0,0 +1,57 @@
...
+                    details(error.detail)
+                    severity(Severity.WARNING)
+                    solution("Run ktlintFormat to auto-fix this issue")
+                }
+            }
+        }
+    }
+
+    fun reportProblem(error: SerializableLintError, filePath: String) {
+        val reporter: ProblemReporter? = problems?.reporter
```
reinsch82, line 45: Can you reuse this in the above function?

vanessamj99:
Hi, could I get some eyes on this pr? Thanks in advance! @Tapchicoma @JLLeitschuh @wakingrufus

plugin/build.gradle.kts
```diff
@@ -118,6 +118,7 @@ dependencies {
     compileOnly(libs.kotlin.gradle.plugin)
     compileOnly(libs.android.gradle.plugin)
     compileOnly(kotlin("stdlib-jdk8"))
+    compileOnly(gradleApi())
```
wakingrufus, line 121: this should not be needed, as we have the `kotlin-dsl` plugin already

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/ConsoleReportWorkAction.kt
```diff
@@ -42,6 +44,15 @@ internal abstract class ConsoleReportWorkAction : WorkAction<ConsoleReportWorkAc
         }
 
         val isLintErrorsFound = lintErrors.values.flatten().isNotEmpty()
+        // The gradle version needs to be at least 8.14 to use the problems API
+        if (isLintErrorsFound && GradleVersion.current() > GradleVersion.version("8.14")) {
+            try {
+                val problemsApiReporter = ProblemsApiReporter()
+                problemsApiReporter.reportProblems(lintErrors)
```
JLLeitschuh, line 51: I'm confused, isn't this a no-op because `problems` will always be `null`?

JLLeitschuh (commented):
I'm finding myself confused, it's not clear to me how the `ProblemsApiReporter` is actually used in the codebase in this PR. The `ProblemsApiReporter` is never instantiated via dependency injection, and doesn't seem to be properly wired into Gradle's problems API in any meaningful way.

Am I missing something? I'm finding myself confused here.

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/reporter/ProblemsApiReporter.kt
```diff
@@ -0,0 +1,38 @@
...
+            errors.forEach { error ->
+                reportProblem(error, filePath)
+            }
+        }
+    }
+
+    fun reportProblem(error: SerializableLintError, filePath: String) {
+        val reporter: ProblemReporter = problems.reporter
+
+        val group = ProblemGroup.create("validation", "ktlint issue")
```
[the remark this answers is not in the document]
JLLeitschuh: Also... Looking at other uses of `ProblemGroup.create` on GitHub, I don't think the the group `"validation"` makes sense.

https://github.com/search?q=ProblemGroup.create+%28language%3AKotlin+OR+language%3AJava%29&type=code

I'm thinking like `"ktlint-gradle"` makes more sense as a group id and something like `"ktlint-gradle issue"` is a better display name?

Thoughts @wakingrufus?

oleg-nenashev (commented):
* Please make sure to modify https://github.com/JLLeitschuh/ktlint-gradle/blob/main/CHANGELOG.md 
* It would be also nice to have a documentation entry for Problems API integration. Maybe, the best location is after the https://github.com/JLLeitschuh/ktlint-gradle?tab=readme-ov-file#baseline-support section

CHANGELOG.md
JLLeitschuh, line 1: These changes don't make sense. Can you fix this up?

JLLeitschuh:
Overall, I'm finding myself quite confused about these changes and the back-and-forth we had to engage with on this PR. The vibe I was getting for many of the changes made in this PR were that the code generated was coming from an AI. The original set of changes didn't seem to understand the nuance of what was actually being implemented, nor following the repositories norms around testing things. The `CHANGELOG` changes don't actually properly flag what was changed. The original set of changes didn't even function properly, and were testing logic in such a way that was self-validating rather than actually validating the changes were implementing the designated feature. I'm happy to educate someone on how to implement features in this repository, but this whole PR review process felt weird and outside of the norm I'm used to experiencing, so I'm trying to understand what was actually going on here.

Overall feedback: in the future, if using AI, please utilize critical thinking around the changes proposed to ensure that the changes are actually implementing the feature desired.

vanessamj99:
The reason I used mockito is because that is the testing framework I usually used but I see I should have used the testing already done in this repo. I was implementing this in a different repo but is different here. Will fix the changelog and README.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+Problems API Integration [#927](https://github.com/JLLeitschuh/ktlint-gradle/pull/927)
```
JLLeitschuh, line 10: This should be a list element
```suggestion
 - Problems API Integration [#927](https://github.com/JLLeitschuh/ktlint-gradle/pull/927)
```

JLLeitschuh:
This PR has conflicts with `main`, can you please rebase the changes?

oleg-nenashev:
@JLLeitschuh it was merged

PR #937, opened 2025-08-12
Gradle 9 Support

wakingrufus:
Gradle 9 Support
remove use of ConfigureUtil
update max kotlin version for testing to 2.2.0

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
@@ -1,6 +1,5 @@
 package org.jlleitschuh.gradle.ktlint
 
-import groovy.lang.Closure
 import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.file.ConfigurableFileTree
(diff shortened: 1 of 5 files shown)


plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
```diff
@@ -176,10 +173,8 @@ internal constructor(
         /**
          * Add 3rd party reporters.
          */
-        fun customReporters(configuration: Closure<NamedDomainObjectContainer<CustomReporter>>) {
-            // This method is needed for Groovy interop
-            // See https://discuss.gradle.org/t/multi-level-dsl-for-plugin-extension/19029/16
-            ConfigureUtil.configure(configuration, customReporters)
+        fun customReporters(configuration: Action<NamedDomainObjectContainer<CustomReporter>>) {
```
JLLeitschuh, line 176: Is this a binary API breaking change? IE. if projects are depending upon this plugin as a direct dependency rather than as a script, anything compiled against our APIs won't work anymore, correct?

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/android/KtlintPluginAndroidTest.kt
```diff
@@ -88,11 +88,6 @@ class KtlintPluginAndroidTest : AbstractPluginTest() {
             minimumJava = 11,
             maximumJava = 17
         ),
-        MAX_GRADLE_MIN_AGP(
-            GradleVersion.version(TestVersions.maxSupportedGradleVersion),
-            TestVersions.minAgpVersion,
-            TestVersions.minSupportedKotlinPluginVersion
-        ),
```
JLLeitschuh, line 95: Why drop this?

JLLeitschuh (commented):
A few questions

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/android/KtlintPluginAndroidTest.kt
wakingrufus: the old AGP doesnt work with Gradle 9, so this combination no longer works

plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt
wakingrufus: possibly... I know gradle has some stuff to auto convert between SAMs and Closures, but I'm not sure if it would help here

wakingrufus: in any case, this is the way is is supposed to be done in Gradle now, so I think at most, I'll make a note about it in release notes

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/testdsl/TestAnnotations.kt
```diff
@@ -25,7 +29,6 @@ object TestVersions {
 @Retention(AnnotationRetention.RUNTIME)
 annotation class GradleTestVersions(
     val minVersion: String = TestVersions.minSupportedGradleVersion,
-    val maxVersion: String = TestVersions.maxSupportedGradleVersion,
```
wakingrufus, line 28: we were never overriding this anyway, and making it an annotation property was preventing me from making it dynamic based Java version

pkubowicz:
What problems does this PR solve?

I've been using ktlint-gradle 13.0.0 with Gradle 9.0.0 for more than two weeks, without any problems. When I execute `gradle --warning-mode=all ktlintCheck ktlintFormat`, I don't receive any warnings.

Issue #942, opened 2025-08-18
Add Rule Version Filtering Based on @SinceKtlint Annotations


hugoncosta:
# Enhancement: Add Rule Version Filtering Based on @SinceKtlint Annotations

## Expected Behavior

Add a new configuration option to KtLint plugin that allows users to filter out rules that were introduced after a specified KtLint version. This would use the existing `@SinceKtlint` annotations present on all rules to determine which rules to include.

**Proposed Configuration:**
```kotlin
ktlint {
    version.set("1.5.0")
    maxRuleVersion.set("0.46.0") // Only include rules introduced up to version 0.46.0
}
```

Why do I want this?
Upgrading to higher versions of Kotlin (e.g. 2.1.0) requires a newer version of Ktlint (we're currently on 1.2.1). If I upgrade it, I will cause failures due to the new rules not being respected, which I'd like to avoid. 

With the proposed feature, I would be able to upgrade to the latest version (e.g. 1.7.1) which _does_ have Kotlin 2.2 compatibility without having to introduce all the new rules between 1.2.1 and 1.7.1. This would then allow me to upgrade the rule set on my own time.

## Current Behavior

No filtering allowed unless I manually exclude new rules myself.

## Additional Information

* Current version of ktlint: all

I initially proposed this in https://github.com/pinterest/ktlint/issues/3099 but was redirected to the plugin itself (you can see the discussion in the thread). Let me know your thoughts on this.

hugoncosta:
No, I don't want any jar shennenigans exactly because of that. I actually tried that internally and it caused a very strange issue where if you were bringing in Ktlint via a third party package (we have team Gradle plugins), it would fail because the logger was broken somehow.

My change is to allow users to define which rules they'd like to run based on the version. The idea here is that I can continously upgrade Ktlint, Kotlin, Gradle without ever having to touch my code until I'm ready to include the new rules being introduced in each version.

> what does this achieve that can't be done with .editorconfig to disable unwanted rules?

I'd need to maintain a list of unwanted rules, which I really do not want to do.

For context, my use case is the following - we maintain a couple of presets, basically an opinionated Gradle setup for all Kotlin packages within the company. I cannot simply force everyone to add the missing commas all at once, but I also cannot be stuck because Kotlin 2.1.0 is incompatible with Ktlint 1.2.1. So my proposed solve is let me upgrade to Ktlint 1.2 and ensure I don't enable any rule that was added past Ktlint 1.2

[...]

PR #943, opened 2025-08-18
Add maxRuleVersion configuration for version-based rule filtering

hugoncosta:
Address the feature request in #942

## Summary
Adds `maxRuleVersion` configuration to allow filtering rules based on their `@SinceKtlint` version annotation.

## Changes
- Add `maxRuleVersion` property to `KtlintExtension`
- Implement version compatibility checking with semantic versioning support
- Filter rules in `KtLintInvocation100` based on `maxRuleVersion` setting
- Add comprehensive tests and documentation

More context in the issue mentioned above.

### DIFF
--- a/plugin/gradle/libs.versions.toml
+++ b/plugin/gradle/libs.versions.toml
@@ -1,6 +1,7 @@
 [versions]
 kotlin = "1.5.31"
 ktlint = "1.0.0"
+ktlintTest = "1.8.0"
 androidPlugin = "4.1.0"
 semver = "1.1.1"
 jgit = "5.13.3.202401111512-r"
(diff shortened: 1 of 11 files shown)


hugoncosta:
For reference, the equivalent change in Ktlint has been merged - https://github.com/pinterest/ktlint/pull/3101

wakingrufus:
Thanks for the contribution!
I understand now that this feature request is to delegate to a new feature of ktlint itself, not something that would effect the dependencies we are setting up. This looks good to me.

wakingrufus:
Looking good. one more question: how will this interact with versions of ktlint from before the annotation was retained at runtime? I think it should fail "open" in that case, and just include all rules. We support using ktlint versions back to 1.0.0, so this must be considered.

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
```diff
@@ -716,4 +716,37 @@ dependencies {
...
+    @CommonTest
+    fun workWithoutMaxRuleVersion(gradleVersion: GradleVersion) {
+        project(gradleVersion) {
+            withCleanSources()
+
+            build(CHECK_PARENT_TASK_NAME) {
+                assertThat(task(":$mainSourceSetCheckTaskName")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+            }
+        }
+    }
```
JLLeitschuh, line 751: So... These test that the `maxRuleVersion` can be set, but not that it behaves in a way that is expected. We should have a test for that.

plugin/src/adapter100/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintInvocation100.kt
```diff
@@ -34,11 +39,62 @@ class KtLintInvocation100(
...
+            val sinceAnnotation = ruleClass.getAnnotation(SinceKtlint::class.java)
+
+            return if (sinceAnnotation != null) {
+                isVersionCompatible(sinceAnnotation.version, maxVersion)
+            } else {
+                // If no annotation, assume it's from an older ktlint version (before annotations were included
+                // at runtime) and include it for backward compatibility
+                true
+            }
+        }
```
JLLeitschuh, line 73: This filtering isn't a feature built into ktlint itself as a library? I guess we are doing the rule loading manually, so I suppose this makes sense.

JLLeitschuh (changes requested):
The tests are the big thing. They need to cover behavior, not just that the flags can/can't be set

plugin/src/test/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPluginTest.kt
hugoncosta: Totally agree, but for that, I think we'll need to park this PR until Ktlint 1.8.0 gets published, as the annotation isn't present at runtime (what we need) before. Or do you know of any other way to build against the unpublished Ktlint version?

plugin/gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 kotlin = "1.5.31"
-ktlint = "1.0.0"
+ktlint = "1.8.0"
```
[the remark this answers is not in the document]
hugoncosta: That's a good idea. I've gone ahead and implemented it using reflection, it's working as expected and doesn't force us to make a new major version :)

hugoncosta:
Hello, anything I can address in this to get it through? We have been using this internally in 1000s of packages successfully.

PR #962, opened 2025-09-22
recreate project isolation failure

wakingrufus:


plugin/settings.gradle.kts
```diff
@@ -8,15 +8,24 @@ pluginManagement {
...
 }
 
 plugins {
     id("com.gradle.develocity") version "3.17"
-    id("org.gradle.toolchains.foojay-resolver-convention") version "0.10.0"
+    id("org.gradle.toolchains.foojay-resolver-convention") version
+        if (
+            Runtime::class.java
+                .let { it.getPackage().specificationVersion ?: it.getMethod("version").invoke(null).toString() }
+                .split(".")[0].toInt() <= 11
```
JLLeitschuh, line 23: Good lord that's messy 😂

wakingrufus: yeah, declaring a separate function doesn't work because the plugin block is evaluated before the rest of the script.

wakingrufus:
@JLLeitschuh would you mind updating the "required checks" to reflect the new action task names?

JLLeitschuh (commented):
Some questions

settings.gradle.kts
```diff
@@ -1,8 +1,6 @@
 import java.util.Properties
 
 pluginManagement {
-    includeBuild("./plugin")
```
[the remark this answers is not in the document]
wakingrufus: its only needed for settings plugins, but I can move it back here, since it is shown this way in the Gradle docs.

.github/workflows/build-and-check.yml
```diff
@@ -83,6 +85,45 @@ jobs:
...
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        with:
+          path: |
+            plugin/.gradle-test-kit/caches/modules-2
+            plugin/.gradle-test-kit/caches/files-2.1
+            plugin/.gradle-test-kit/caches/metadata-2.96
+          key: gradle-wrapper-${{ hashFiles('**/gradlew') }}-${{ github.sha }}
+          restore-keys: |
+            gradle-wrapper-${{ hashFiles('**/gradlew') }}-
+            gradle-wrapper-
```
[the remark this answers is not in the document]
wakingrufus: no this is copy-pasted from another spot, b/c i had to make a separate task

JLLeitschuh:
> would you mind updating the "required checks" to reflect the new action task names?

That's probably going to need to happen post-merge, I can always bypass the rules to let a merge happen

plugin/build.gradle.kts
```diff
@@ -154,19 +172,14 @@ tasks.whenTaskAdded {
     }
 }
 
-// Disabling default jar task as it is overridden by shadowJar
-tasks.named("jar").configure {
-    enabled = false
-}
-
 val ensureDependenciesAreInlined by tasks.registering {
```
JLLeitschuh: Is this task even doing what we were thinking it was doing? Is it doing it adequately?

You said that our pom and gradle metadata dependencies were being exposed as direct dependencies, right?

wakingrufus: I played around with it a bit, I there was nothing I could do to actually make this thing fail. I think the shadow plugin has improved to where it is harder to mess this stuff up now. So I have removed this task

.github/workflows/build-and-check.yml
```diff
@@ -49,7 +49,9 @@ jobs:
       - name: set up JDK
         uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v5.0.0
         with:
-          java-version: ${{ matrix.java_version }}
+          java-version: |
+            ${{ matrix.java_version }}
+            21
```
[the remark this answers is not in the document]
wakingrufus: the daemon toolchain is locked to 21 now, so we always need 21 for gradle itself. the test process is forked off of that using a project toolchain launcher, determined by the parameter we pass in (which is sourced from the matrix)

wakingrufus:
> Is this good to merge? Once I do so, I'll update the GitHub actions required cheks

yep! Thanks

PR #964, opened 2025-09-24
collapse remaining adapter source sets into main

wakingrufus:
since we now only support ktlint 1.0+ we no longer need adapter source sets for older versions, so the adapter source sets for 1.0+ can be merged into main in order to simplify the build

update tests to run against latest gradle and kotlin releases


JLLeitschuh:
IMHO, this seems more annoying to recreate if we need to in the future because of API compatibility issues. I foresee it's more likely that that will happen in the future honestly. Thoughts?

wakingrufus:
I feel that ktlint has made good on the promise of a stable API post-1.0. the separate source set solution was really because the breaking changes were becoming too much to manage with just reflection. In the future, minor breakages can be handled with just reflection. If ktlint 2.0 ever happens with major changes, I'd rather reinstate the source set then have to deal with it the whole time until then. Also, the abstractions needed will probably be different anyway, so it would probably need to be reworked anyway.

JLLeitschuh:
> the breaking changes were becoming too much to manage with just reflection.

Yea... Let's not go back to those days 😂

> If ktlint 2.0 ever happens with major changes, I'd rather reinstate the source set then have to deal with it the whole time until then. Also, the abstractions needed will probably be different anyway, so it would probably need to be reworked anyway.

Alright, this is mostly your ship at this point, I'm mostly just helping navigate. I trust your judgement here.

JLLeitschuh (approved):
LGTM! :shipit:

Issue #1008, opened 2025-12-01
Add support for AGP 9 with built-in Kotlin


mannodermaus:
Adapted from [the Android docs page](https://developer.android.com/build/migrate-to-built-in-kotlin):

> Android Gradle plugin 9.0 introduces built-in Kotlin support and enables it by default. Developers no longer have to apply the `org.jetbrains.kotlin.android` (or `kotlin-android`) plugin in their build files to compile Kotlin source files.

When a project uses built-in Kotlin support, ktlint-gradle won't find a compatible plugin and therefore does not create any of the `ktlintXXSourceSet` tasks necessary to perform the actual checking/formatting of code.

Most likely, the [list of detected plugins](https://github.com/JLLeitschuh/ktlint-gradle/blob/b7686ad49a3d097c754e9b9541b5c18482604790/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt#L30-L36) will need to be updated. From a cursory search of the AGP 9 codebase, there seems to be a `com.android.built-in-kotlin` plugin ID, but it's not applied by default when the setting is enabled. Further investigation needs to be done to figure out how to identify an Android project with built-in Kotlin turned on.

wakingrufus:
> [@wakingrufus](https://github.com/wakingrufus) You might want to refer to this section of code:
> 
> https://github.com/Goooler/agp-sources/blob/8c31acc72d0db3dbdb45e297f98add082b49bcf9/9.0.0-rc02/com.android.tools.build/gradle/com/android/build/gradle/internal/services/BuiltInKotlinServices.kt#L232-L258
> 
> It shows how the built-in Kotlin is applied.

Based on this, I think we _should_ use the id, not class loading. @mannodermaus would you mind updating your PR?

PR #1016, opened 2025-12-18
Add support for Android projects with new DSL i.e. built-in Kotlin

mannodermaus:
Android Gradle Plugin 9.0 introduces support for Android projects using "built-in Kotlin" ([reference](https://developer.android.com/build/migrate-to-built-in-kotlin)), a new way of integration between the AGP and Kotlin. The current Ktlint Gradle plugin doesn't recognize projects with this setup, because these projects don't rely on the `org.jetbrains.kotlin.android` plugin. Instead, a common way to approach this is to hook into the pairing of the `com.android.base` and `KotlinBaseApi` plugins ([here](https://github.com/JakeWharton/cite/blob/685896406588e8ee0057491d7d67add443da6938/cite-gradle-plugin/src/main/kotlin/com/jakewharton/cite/plugin/gradle/CitePlugin.kt#L57-L58) is an example).

Resolves https://github.com/JLLeitschuh/ktlint-gradle/issues/1008.

### DIFF
--- a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
+++ b/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintPlugin.kt
@@ -51,6 +51,8 @@ open class KtlintPlugin : Plugin<Project> {
     @Suppress("UNCHECKED_CAST")
     private val kotlinBaseApiPluginClass by lazy {
         try {
+            // TODO: After raising the baseline Kotlin version to 1.7.20 or newer,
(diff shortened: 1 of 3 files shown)


JLLeitschuh (approved):
This looks fine to me, but I really want feedback from @wakingrufus

mannodermaus:
No worries at all! I hope you had a relaxing time over the holidays. 😊 I added a changelog entry as requested.

wakingrufus:
> @wakingrufus can we also do a release ? 🫣 😁 

Yeah I'll try to get to this tomorrow

PR #1027, opened 2026-01-19
Some code modernization+deprecation fixes+version upgrades

wertzuz:
mainly for the samples. Hope it's all fine if you see something additional i missed or have an issue with one change i'm happy to update this pr accordingly.

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -40,17 +40,18 @@ ktlint {
 
 tasks.withType<KotlinCompile> {
     compilerOptions {
-        languageVersion.set(KotlinVersion.KOTLIN_1_8)
-        apiVersion.set(KotlinVersion.KOTLIN_1_8)
-        jvmTarget.set(JvmTarget.JVM_1_8)
+        languageVersion.set(KotlinVersion.KOTLIN_2_1)
+        apiVersion.set(KotlinVersion.KOTLIN_2_1)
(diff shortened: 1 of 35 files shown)


JLLeitschuh (approved):
Hey @wertzuz! Thanks so much for this contribution!

Sorry it's taken so long for us to get back to you, I've been busy with life.

I'm happy to merge this, but I'd love a review pass from @wakingrufus really quickly.

Ideally this would have a change log entry associated with it, but it may not be really necessary as it's not a functional change.

If you add a CHANGELOG.md entry, it will unblock the CI pipe on this

wakingrufus (changes requested):
We compile against ktlint 1.0 because we support using that version of ktlint still, and use parameterized tests to ensure newer versions work

wertzuz:
I consolidated versions and moved ktlint back to 1.0.1 since some targets had already .1 set. Also could be worth thinking about bumping the ktlint minimum required. @wakingrufus @JLLeitschuh

plugin/build.gradle.kts
```diff
@@ -97,7 +97,7 @@ testing {
         named<JvmTestSuite>("test") {
             useJUnitJupiter()
         }
-        listOf(8, 11, 17, 21, 25)
+        listOf(11, 17, 21, 25)
```
JLLeitschuh, line 100: Why are we dropping java 8 unit testing here?

PR #1059, opened 2026-06-01
Kotlin 2.4 support

wakingrufus:
No implementation changes were needed; I just updated tests to prove support

test against kotlin 2.4
don't test with deprecated js plugin in kotlin 2.4+ 
update examples to test against kotlin 2.4.0


JLLeitschuh (commented):
LGTM. I'd prefer to get us off a beta tag as soon as possible. But I'm fine with this

JLLeitschuh (approved):

Issue #893, opened 2025-05-28
Clean up workaround for Gradle 6 in Configurations.kt


tadjan:
Since the minimal supported Gradle version is now 7.4, the workaround for Gradle 6 in [plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt](https://github.com/JLLeitschuh/ktlint-gradle/blob/15c739266dc63a0ef03653d78683f3d3e7f54ead/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt#L35) can now be removed. 
```
        // Workaround for gradle 6 https://github.com/gradle/gradle/issues/13255
        val oldProp = target.objects.listProperty(Dependency::class.java)
        dependencies.addAllLater(
            oldProp.value(
                extension.version.map {
                    if (SemVer.parse(it) < SemVer(1, 0, 0)) {
                        target.logger.info("Add dependency: ktlint version $it")
                        listOf(target.dependencies.create("com.pinterest:ktlint:$it"))
                    } else {
                        target.logger.info("Add dependencies: ktlint version $it")
                        listOf(
                            target.dependencies.create("com.pinterest.ktlint:ktlint-cli:$it"),
                            // these transitive deps were introduced in ktlint 1.0, but for some reason not picked up automatically
                            target.dependencies.create("io.github.oshai:kotlin-logging:5.1.0"),
                            target.dependencies.create("io.github.detekt.sarif4k:sarif4k:0.5.0")
                        )
                    }
                }
            )
        )
```
It's not only obsolete code at this point, but also hardcodes dependencies on old versions of `kotlin-logging` and `sarif4k` libraries.

PR #1068, opened 2026-06-06
Clean up workaround for Gradle 6

tadjan:
PR for issue https://github.com/JLLeitschuh/ktlint-gradle/issues/893

Since the minimal supported Gradle version is now 7.4, the workaround for Gradle 6 in [plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/Configurations.kt] can now be removed.

It's not only obsolete code at this point, but also hardcodes dependencies on old versions of kotlin-logging and sarif4k libraries.


JLLeitschuh (approved):
LGTM!

JLLeitschuh:
@wakingrufus any concerns? Otherwise, I'm happy to see this merged

wakingrufus (approved):

PR #1112, opened 2026-08-12
Propagate JAVA_TOOL_OPTIONS to the KtLint Worker Daemon process

### LINKED ISSUE #1110: Propagate environment config
I am having to run ktlint under a tool that runs in (macos) a sandbox.

In order to be able to do this, and not have the sandbox break, I find I am having to pass an environment variable. I have to force the interaction to use ipv4 because the sandbox does not cope well with ipv6. I am concurrently working to fix the sandbox.

I would like to have the opportunity to pass an environment variable to make things work out. The patch I suggest is something like this:

```kotlin
private fun submitKtLintWork(
    filesToCheck: Set<File>,
    formatSources: Boolean,
    editorConfigUpdated: Boolean,
    formatSnapshot: File? = null
) {
    val queue = workerExecutor.processIsolation {
        classpath.from(ktLintClasspath, ruleSetsClasspath)
        forkOptions {
            maxHeapSize = workerMaxHeapSize.get()

            if (JavaVersion.current() >= JavaVersion.VERSION_16) {
                jvmArgs("--add-opens", "java.base/java.lang=ALL-UNNAMED")
            }

            // Gradle strips JAVA_TOOL_OPTIONS from forwarded worker environments by
            // default; re-propagate it explicitly so JVM flags injected via this env
            // var (used by some sandboxing/monitoring tooling) remain active for the
            // Worker Daemon, matching the behavior of Test task executors.
            System.getenv("JAVA_TOOL_OPTIONS")?.let { javaToolOptions ->
                environment("JAVA_TOOL_OPTIONS", javaToolOptions)
            }
        }
    }
    ...
}
```

This would allow me to pass ` -Djava.net.preferIPv4Stack=true` to work around my sandbox issues. I have tested this locally, and it removes my ktlint issues.

geirh-nav:
## Summary

Fixes #1110

Gradle does not forward the `JAVA_TOOL_OPTIONS` environment variable to forked worker processes by default, unlike other forked JVMs such as `Test` task executors. This breaks tooling that relies on JVM flags injected via `JAVA_TOOL_OPTIONS` — for example, sandboxing/monitoring tools that require flags like `-Djava.net.preferIPv4Stack=true` to be active in every forked JVM, not just the main Gradle process or `Test` workers.

This PR re-propagates `JAVA_TOOL_OPTIONS` explicitly into the KtLint Worker Daemon's process environment when it is present in the environment running Gradle, matching the behavior of `Test` task executors. This mirrors the workaround suggested in the issue and verified there by the reporter.

## Changes

- `BaseKtLintCheckTask`: forward `JAVA_TOOL_OPTIONS` from the Gradle process environment into the Worker Daemon's `forkOptions` when set.
- `README.md`: document the new automatic behavior.
- `CHANGELOG.md`: add an entry under `[Unreleased]`.
- `KtlintPluginTest`: add a functional test that sets `JAVA_TOOL_OPTIONS` via `GradleRunner.withEnvironment` and asserts the JVM's "Picked up JAVA_TOOL_OPTIONS" startup diagnostic appears in the forked worker's output, confirming propagation reaches the Worker Daemon (not just the Gradle daemon).

## Testing

- `:plugin:compileKotlin` / `:plugin:compileTestKotlin` succeed.
- The new functional test was run locally and printed `Picked up JAVA_TOOL_OPTIONS: -Dktlint.gradle.test.marker=true` from the forked KtLint Worker Daemon process, confirming the fix works as intended.
- Note: in my local sandboxed environment, loopback TCP sockets are blocked, which prevents Gradle's Worker Daemon IPC from completing (this reproduces identically on unmodified `main`, so it is unrelated to this change). Full test suite execution should be verified via this repository's CI.


JLLeitschuh:
This seems like a bug in Gradle, not in downstream plugins. Do you have an issue open against http://github.com/gradle/gradle?

geirh-nav:
Good question — no, there wasn't one yet. Filed it now: https://github.com/gradle/gradle/issues/38836

geirh-nav:
I actually prefer the fix-it-in-gradle idea, so I'll make a PR for gradle to sort this instead. If they don't want my change I might have to add another PR for this.

